PR #5, opened 2019-07-25
Init Percona helm repo

cap1984:


paulczar:
I would recommend against this!

keeping the chart .tgz files in the repo is a smell.

The `chart-releaser` tool will create github releases for the charts themselves and update the `index.yaml` to reflect.  

The circleci tasks here - https://github.com/Percona-Lab/percona-helm-charts/blob/master/.circleci/config.yml will actually do all of this for you with testing etc enabled.  If you want to use something other than circle the scripts etc it calls should still work.

For running the steps manually, the second half of this blog post describes what should be done:

https://tech.paulcz.net/blog/creating-a-helm-chart-monorepo-part-1/

cap1984:
> I would recommend against this!
> 
> keeping the chart .tgz files in the repo is a smell.
> 
> The `chart-releaser` tool will create github releases for the charts themselves and update the `index.yaml` to reflect.
> 
> The circleci tasks here - https://github.com/Percona-Lab/percona-helm-charts/blob/master/.circleci/config.yml will actually do all of this for you with testing etc enabled. If you want to use something other than circle the scripts etc it calls should still work.
> 
> For running the steps manually, the second half of this blog post describes what should be done:
> 
> https://tech.paulcz.net/blog/creating-a-helm-chart-monorepo-part-1/

I've followed your CI steps, and the release is available now. 
I'm updating index.yaml in order to push those changes into the GitHub pages

paulczar:
looks great!

PR #6, opened 2019-08-01
update readme with correct url for percona-db chart.

paulczar:
Signed-off-by: Paul Czarkowski <username.taken@gmail.com>


paulczar:
Also there's automation in the CIRCLE scripts to keep the README and other docs in `gh-pages` branch up to date.   That's not running so they're not quite right.

see https://github.com/Percona-Lab/percona-helm-charts/blob/master/.circleci/release.sh#L90-L104

PR #7, opened 2019-09-16
support cluster wide operation

kubicgruenfeld:
this is basically a copy of https://github.com/paulczar/percona-helm-charts/pull/1, but to the correct upstream repository.

with this change it is possible to watch all namespaces for percona-xtradb-cluster deployments.

### DIFF
--- a/charts/pxc-operator/templates/deployment.yaml
+++ b/charts/pxc-operator/templates/deployment.yaml
@@ -31,7 +31,11 @@ spec:
           - percona-xtradb-cluster-operator
           env:
             - name: WATCH_NAMESPACE
-              value: {{ default .Release.Namespace .Values.watchNamespace }}
+            {{- if kindIs "invalid" .Values.watchNamespace }}
+              value: {{ .Release.Namespace }}
+            {{- else }}
+              value:  {{ .Values.watchNamespace }}
+            {{- end }}
             - name: OPERATOR_NAME
               value: {{ default "percona-xtradb-cluster-operator" .Values.operatorName }}
           # livenessProbe:
--- a/charts/pxc-operator/templates/role-binding.yaml
+++ b/charts/pxc-operator/templates/role-binding.yaml
@@ -1,4 +1,8 @@
+{{- if kindIs "invalid" .Values.watchNamespace }}
 kind: RoleBinding
+  {{- else }}
+kind: ClusterRoleBinding
+{{- end }}
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
   name: {{ include "pxc-operator.fullname" . }}
@@ -10,14 +14,14 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: default
-  {{- if .Values.watchNamespace }}
+  {{- if not (kindIs "invalid" .Values.watchNamespace) }}
   namespace: {{ .Release.Namespace }}
   {{- end }}
 roleRef:
-  {{- if .Values.watchNamespace }}
-  kind: ClusterRole
-  {{- else }}
+  {{- if kindIs "invalid" .Values.watchNamespace }}
   kind: Role
+  {{- else }}
+  kind: ClusterRole
   {{- end }}
   name: {{ include "pxc-operator.fullname" . }}
   apiGroup: rbac.authorization.k8s.io
--- a/charts/pxc-operator/templates/role.yaml
+++ b/charts/pxc-operator/templates/role.yaml
@@ -1,7 +1,7 @@
-{{- if .Values.watchNamespace }}
-kind: ClusterRole
-{{- else }}
+{{ if kindIs "invalid" .Values.watchNamespace }}
 kind: Role
+{{- else }}
+kind: ClusterRole
 {{- end }}
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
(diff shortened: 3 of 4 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/Percona-Lab/percona-helm-charts?pullRequest=7) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/Percona-Lab/percona-helm-charts?pullRequest=7) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/Percona-Lab/percona-helm-charts?pullRequest=7) it.</sub>

delgod:
@kubicgruenfeld 
we think that particularly "database operators" should work in a single namespace.
yes, it sounds weird, but the reason is - database is a complex stateful application.
we really do our best to get fearless experience during database upgrades.
due to this our operator support only two versions - current and the previous.
so, if you have many namespaces
1) it is hard to have the same database version in all databases
2) we don't want to have a situation when operator update for namespace "test1" can (even theoretically) affect database in namespace "prod"

stefanandres:
This is sad for us. We are using several 3rd party operators in our clusters (nginx-ingress, postgres operator from zalando, presslabs mysql-operator).

We'd like to switch from presslabs to percona-operator because we have had a good experience running >500 servers with percona.

To 1.: The database version should not be defined by the operator itself, but by the object you feed the operator with. E.g. you can specify what database version you want in the [postgresql object](https://postgres-operator.readthedocs.io/en/latest/user/#create-a-manifest-for-a-new-postgresql-cluster). So the operator must have a matrix of supported db versions.

To 2.: Our kubernetes cluster are already seperated by environment, but each namespace represents an entire application. I.e. when running a setup, a namespace includes 1. a db 2. some caching 3. an ingress object 4. monitoring objects. 5. application pods. This makes is easy to implement namespace network policies and namespace resource quotes.

Running operators cluster-wide seems to be the common best practice in the community currently.

https://kubernetes.io/docs/concepts/extend-kubernetes/operator/ references to https://cloud.google.com/blog/products/containers-kubernetes/best-practices-for-building-kubernetes-operators-and-stateful-apps guideline which references to accepted https://github.com/kubernetes-sigs/kubebuilder framework. This framework uses cluster-wide [concepts](https://github.com/kubernetes-sigs/kubebuilder/search?p=1&q=kind%3A+role&type=Code&unscoped_q=kind%3A+role)

tduzan:
Howdy @stefanandres,

I understand your use case, and it is possible for us to build the operator so it supports cluster-wide operation.  From a development perspective this is a pretty straightforward change, but from an operational and user experience perspective it comes with a lot of caveats.  Our initial decision was based on a concern of handing users inexperienced with Kubernetes what amounts to a loaded gun pointed at their foot.  

Upgrades for most organizations largely are not optional, in order to meet security and compliance regulations, and upgrades should be a safe operation.  It's probably the area of database operations most fraught, and we wanted to use Operators as an opportunity to make this something that's safe by default.  Restricting the Operator to a single-namespace helps us to ensure this by making it possible for a user to run different Operator and database versions in different namespaces easily, in a supported way.

For us, "support" means more than just "the software allows this to happen", it means we very much support our users doing this type of configuration and will assist to make it work.  Given that, we have to be somewhat pragmatic when limiting the size of our support matrix, as we do thorough testing and validation at every level of the matrix.  In this vein, we made the choice that each Operator release would support images intended for itself and its immediate prior release, but no earlier.  Nothing prevents someone from running images several releases back with a newer Operator, and in theory it'd work fine, but we cannot easily invest the resources needed to validate to the level of thoroughness we do with all our releases that many versions.

By using separate Operators per namespace, a user can therefore run a supported release, at any point in the release history, separate from the remaining environment.  Essentially, we chose that the namespace would be our environmental isolation boundary, and you're saying that in your environment rather than having a dev and prod namespace, you'd have dev and prod clusters so the cluster is your environmental isolation boundary and namespaces are per-application.

We'll discuss internally and identify if there's a clean way to support both methods, still defaulting to Operator per namespace, and see what we can do.  Generally, my goal with our Operators is that we provide a pathway to do anything reasonable which would be supported within Kubernetes for advanced users, but our defaults should always help users who are unfamiliar with Kubernetes to have a good experience using our Operators with a mind towards necessary configuration for production workloads.  It's our goal to make production databases on Kubernetes not just a thing, but something that is admirable.

Hope this helps to explain the decision mindset, and I will seek to follow-up with you here on Github later this week.

Thanks.

stefanandres:
Hey @tduzan,
thank you for taking your time and explaining your decision more elaborately.
I'm glad you are taking an approach you feel good with to aim for a good and stable product.
We're eager to test and see whether/how we can integrate your operator in our customer experience.

tduzan:
I'm missing some of the folks I need to discuss this with, since they're out at a conference this week.  I will follow-up again next week after I discuss it with them on Monday.

baurmatt:
Hey @delgod / @tduzan, as Percona now seems to [officially support](https://www.percona.com/doc/kubernetes-operator-for-pxc/cluster-wide.html) cluster-wide operator - would you mind taking another look at this PR? :)

PR #8, opened 2020-01-06
CLOUD-481 Chart upgrade due to new version release

cap1984:


charts/pxc-db/Chart.yaml
```diff
@@ -1,9 +1,9 @@
-apiVersion: v1
-appVersion: "1.1.0"
+apiVersion: v1-3-0
```
paulczar: this change of API version is probably going to cause issues, this is the version for the Helm API.  see https://helm.sh/docs/topics/charts/#the-apiversion-field

charts/pxc-db/templates/cluster.yaml
```diff
@@ -1,6 +1,9 @@
-apiVersion: pxc.percona.com/v1
+apiVersion: pxc.percona.com/{{ .Chart.ApiVersion }}
```
paulczar: see previous comment.

charts/pxc-db/templates/cluster.yaml
```diff
@@ -1,6 +1,9 @@
...
 kind: PerconaXtraDBCluster
 metadata:
+  annotations:
+    kubectl.kubernetes.io/last-applied-configuration: |
+        {"apiVersion":"pxc.percona.com/{{ .Chart.ApiVersion }}","kind":"PerconaXtraDBCluster"}
```
paulczar: normally you wouldn't see a last-applied-configuration and let Kubernetes do that itself.   What's the reason for adding this here ?

paulczar:
I'm not 100% sure the comments in my review are the case of the CI failures.

Helm has also changed its recommendation about CRDs, saying to pull them out of templates and put them untemplated in a seperate `crd` directory, I would recommend updating this to do that.

https://helm.sh/docs/topics/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you

charts/pxc-db/templates/cluster.yaml
cap1984: We are facing the following error without the annotations 

`    - lastTransitionTime: "2020-01-20T18:48:17Z"
      message: 'wrong PXC options: set version: new version: Malformed version: '
      reason: ErrorReconcile
      status: "True"
      type: Error`

charts/pxc-db/Chart.yaml
cap1984: done

charts/pxc-db/templates/cluster.yaml
cap1984: done

PR #14, opened 2020-01-25
K8SPXC-219 Helm 3 support

cap1984:


charts/pxc-operator/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v1
 appVersion: "1.3.0"
-description: A Helm chart for Deploying the Percona XtraDB Cluster Operator Kubernetes
+description: A Helm 3 chart for Deploying the Percona XtraDB Cluster Operator Kubernetes
```
paulczar: I don't think we need to specify `Helm 3` in the description.  This chart should continue to work for `Helm 2` as well, with the additional instructions of `kubectl apply -f crd/` beforehand.

paulczar:
You shouldn't need to copy the `crd` to `charts/pxc-db/crds/crd.yaml` ... only the operator chart itself should have the crd.

cap1984:
> You shouldn't need to copy the `crd` to `charts/pxc-db/crds/crd.yaml` ... only the operator chart itself should have the crd.

Unfortunately in case of missing operator, Kubernetes will throw an error regarding PXC object version types

charts/pxc-operator/Chart.yaml
cap1984: ok

Jancis:
We're using namespace limited SA for deployment and since `pxc-db` also wants to install crd, our deployment fails. `pxc-db` crd is identical to `pxc-operator` crd, it's already set up, please don't have it in `pxc-db` (having an operator ir a pre-requisite).

PR #21, opened 2020-07-01
Remove empty .spec.names.shortNames default from backup CRD.

sfynx:
When keeping a YAML manifest in sync using GitOps tools like ArgoCD, these tools report the manifest being out of sync when the live manifest in the cluster differs from the desired manifest in the source Git/Helm repository.

The `PerconaXtraDBBackup` CRD has an explicit empty list for its `.spec.names.shortNames` property. This is actually the default value, and in that case it apparently gets removed from the live manifest by k8s when applied. This causes the manifest to never become synced in tools like ArgoCD.

The fix is to remove the default empty value from the manifest, which will result in the same configuration on the cluster, but this time in sync with the source manifest.

This is how the issue manifests itself in ArgoCD specifically:
Screen Shot 2020-07-01 at 23 45 04

### DIFF
--- a/charts/pxc-db/Chart.yaml
+++ b/charts/pxc-db/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.4.0"
 description: A Helm chart for installing Percona XtraDB Cluster Databases using the PXC Operator.
 name: pxc-db
 home: https://www.percona.com/doc/kubernetes-operator-for-pxc/kubernetes.html
-version: 0.1.9
+version: 0.1.10
 maintainers:
   - name: paulczar
     email: username.taken@gmail.com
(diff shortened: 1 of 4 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/Percona-Lab/percona-helm-charts?pullRequest=21) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/Percona-Lab/percona-helm-charts?pullRequest=21) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/Percona-Lab/percona-helm-charts?pullRequest=21) it.</sub>

tplavcic:
Thanks for the fix! To me this doesn't look problematic, but now I'm thinking that if we merge this then it would be good to make this change here also: https://github.com/percona/percona-xtradb-cluster-operator/blob/master/deploy/crd.yaml#L161
but then I'm not sure if it can break something or not (thinking about upgrades).
@cap1984 @delgod What do you think?

delgod:
> @cap1984 @delgod What do you think?

yes, we need change in both places

PR #24, opened 2020-07-09
Add example for changing endpoint url for s3

appunni-dishq:
s3 interface can be used in multiple cloud providers. So endpointUrl could be useful for anyone using non standard s3 interface of aws. eg: DigitalOcean

### DIFF
--- a/charts/pxc-db/values.yaml
+++ b/charts/pxc-db/values.yaml
@@ -186,6 +186,7 @@ backup:
     #     bucket: S3-BACKUP-BUCKET-NAME-HERE
     #     credentialsSecret: my-cluster-name-backup-s3
     #     region: us-west-2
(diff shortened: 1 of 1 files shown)


appunni-dishq:
https://github.com/percona/percona-xtradb-cluster-operator/blob/8c4c239cffbebf861fe4541a39f484d374ac0122/pkg/apis/pxc/v1/pxc_types.go

```
type BackupStorageS3Spec struct {
	Bucket            string `json:"bucket"`
	CredentialsSecret string `json:"credentialsSecret"`
	Region            string `json:"region,omitempty"`
	EndpointURL       string `json:"endpointUrl,omitempty"`
}
```
I just noticed that this is available in actual definitions. But not provided in examples

tplavcic:
Thank you for the update!

PR #25, opened 2020-07-09
PMM not found (bug) while installing proxysql

appunni-dishq:


tplavcic:
Thank you for the update!

appunni-m:
I don't think PMM is working , I think it is a bug. After enabling PMM it stopped working for me even after doing this. Who ever disabled it did it to disable PMM

tplavcic:
Hi, your fix seems to be fine, if PMM is not working then it's another problem and needs to be solved on another level. Enabling PMM here only enables PMM agents, so PMM needs to be installed and setup before enabling here.

appunni-m:
Okay seems fair. I will bump version next time. Thanks

PR #29, opened 2020-08-13
K8SPXC-343 Readme update

cap1984:


charts/pxc-db/README.md
```diff
@@ -1,16 +1,149 @@
...
+| `haproxy.podDisruptionBudget.maxUnavailable` | Instruct Kubernetes about the failed pods allowed quantity           | `1`                                      |
+| `haproxy.persistence.enabled` | Requests a persistent storage (`hostPath` or `storageClass`) from K8S for HAProxy Pods  | `true`                                      |
+| `haproxy.persistence.hostPath` | Sets datadir path on K8S node for all HAProxy Pods. Available only when `pxc.persistence.enabled: true` |                             |
+| `haproxy.persistence.storageClass` | Sets K8S storageClass name for all HAProxy Pods PVC. Available only when `pxc.persistence.enabled: true` | `-`                      |
+| `haproxy.persistence.accessMode` | Sets K8S persistent storage access policy for all HAProxy Pods | `ReadWriteOnce`                      |
+| `haproxy.persistence.size` | Sets K8S persistent storage size for all HAProxy Pods | `8Gi`                      |
+| `haproxy.persistence.size` | Sets K8S persistent storage size for all HAProxy Pods | `8Gi`                      |
```
tplavcic, line 89: This seems to be listed twice.
Also notice that Mykola says that haproxy should not create pvc's at all - https://jira.percona.com/browse/K8SPXC-400
so probably the behaviour in operator and helm charts will need to be changed.

cap1984: I've removed duplicated entry.
Actually volume settings for haproxy is a result of using the same structure PodSpec for all pod types (haproxy, pxc, proxysql). Maybe will be changed in the future.

PR #32, opened 2020-09-29
Use v1 apiVersion for CRDs when available

endrec:
Signed-off-by: Endre Czirbesz <endre.czirbesz@rungway.com>


tplavcic (changes requested):
There seems to be some issue with this change:
```
Error: failed to install CRD crds/crd.yaml: error parsing : json: line 0: invalid character '{' looking for beginning of object key string
```
Please fix if there is still interest to merge this.
Thanks!

endrec:
I'm not using Percona any more, so I'll close this PR for now.

PR #67, opened 2021-04-10
Finalizers described

SachinHR:
I have described the finalizers in 'README.md' file within pxc-db folder. I have added a three rows in configurable parameters table.


tplavcic:

PR #71, opened 2021-05-04
support securityContext configuration

pauvos:
adds support for container securityContexts, e.g.:

```yaml
        securityContext:
          runAsUser: 1000
          runAsGroup: 2000
          runAsNonRoot: true
```

This already works for the pxc-operator deployment and creates `PerconaXtraDBCluster` objects with correct securityContext.
But the configuration isn't interpreted and applied to pxc and haproxy statefulsets by the operator yet. securityContext is always empty. :/

### DIFF
--- a/charts/pxc-db/production-values.yaml
+++ b/charts/pxc-db/production-values.yaml
@@ -50,6 +50,7 @@ pxc:
       cpu: 600m
     # db resources are sacred, so don't limit it.
     limits: {}
+  securityContext: {}
   nodeSelector: {}
   #  disktype: ssd
   affinity:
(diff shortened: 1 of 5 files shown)


tplavcic:
Hi @pauvos and thanks for the PR.
PXC pods support `podSecurityContext` and `containerSecurityContext` options: https://www.percona.com/doc/kubernetes-operator-for-pxc/operator.html#pxc-containersecuritycontext
So I will close this PR and since I'm making a change for PXC operator release 1.10 I will add these two options and also `securityContext` for operator deployment.

PR #73, opened 2021-05-21
[pxc-db] Add resources to sidecars, pmm and logcollector

pbabilas:
Enable resources managment to avoid duplicate resources in containers.


pbabilas:
Are there some estimation when it can be set into charts?

michaelcoburn:
Hi @tplavcic poke

tplavcic:
Thank you for the contribution and sorry for waiting! It has now been merged, but notice there is some inconsistencies how it works from the operator side so some changes might happen in the future:  https://jira.percona.com/browse/K8SPXC-868

PR #74, opened 2021-05-27
Add support for pxc.sidecars

baurmatt:
Support for PXC pod sidecars have been added in 1.8.0 release of the
operator. This makes it configurable.


tplavcic:
Thank you for the contribution! It has now been merged.

PR #90, opened 2021-08-24
K8SPXC-771 PXC Operator 1.9.0 chart

cap1984:


charts/pxc-operator/Chart.yaml
```diff
@@ -1,9 +1,9 @@
-apiVersion: v1
-appVersion: "1.8.0"
+apiVersion: v2
+appVersion: "1.9.0"
```
tplavcic, line 2: I see in pxc-db we have changed `appVersion` to be without quotes, maybe just make it consistent.

tplavcic (changes requested):

charts/pxc-operator/Chart.yaml
cap1984: done

tplavcic (approved):

PR #93, opened 2021-08-30
Fix backup.serviceAccountName not being interpolated

smaftoul:
It acutally matches the current documentation.


smaftoul:
@tplavcic Hello ! I think I haved signed the CLA but the workflow seems to wait for an approval. Did I make a mistake ? Is there something left to do on my side ?

PR #97, opened 2021-10-08
CLOUD-767 - inclusion of .Release.Namespace in metadata objects

pszymanski-piwik:
Description

In our specific case, we are using helm template to render templates that fully describe our cluster's state in a declarative way, and then we load them to kubernetes using kubectl apply, in a GitOps fashion.
In order for this to work, we need to include information about namespace inside the templates.
One way to do this, is by including

```
metadata:
```

In all namespace-scoped resource definitions. 

My question is: would you accept PR with this functionality added? After some tests our additions did not brake helm's existing behaviour.

### DIFF
--- a/charts/pxc-operator/Chart.yaml
+++ b/charts/pxc-operator/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: 1.9.0
 description: A Helm chart for Deploying the Percona XtraDB Cluster Operator Kubernetes
 name: pxc-operator
 home: https://www.percona.com/doc/kubernetes-operator-for-pxc/kubernetes.html
-version: 1.9.1
+version: 1.9.2
 maintainers:
   - name: cap1984
     email: ivan.pylypenko@percona.com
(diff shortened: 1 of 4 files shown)


dalbani:
FYI: see the (very long) discussion about this topic in <https://github.com/helm/helm/issues/5465>.

qdupuy:
+1 for this feature

spron-in:
Hey @pszymanski-piwik ,

Thank you for the PR. Looks like a good idea and a valid use case.

Two things we will do on our end:

1. If we decide that we want to add it, we will need to change it in other helm charts (to maintain consistency). If you want to help here - would be awsm)
2. We will take some time (2-3 days) now to check if there are any unexpected consequences here. Like affect upgrade process, various distros installation, etc

nmarukovich:
@tplavcic I believe we can close this one.

tplavcic:
@pszymanski-piwik @pszymanski-piwik @qdupuy We are looking at this now and want to add it, but we would like to not 
 change the current default behaviour.
So a question: Would it work for you if we add an option to all the charts like `includeNamespace` which would be default `off` and if you change it to `on` we would add `namespace: {{ .Release.Namespace }}` into generated output?

tplavcic:
Thank you! This change was done under this PR: https://github.com/percona/percona-helm-charts/pull/241

PR #98, opened 2021-10-16
fix: we should used users values when is set

theyough:


tplavcic:
I have merged you PR to my release branch here: https://github.com/percona/percona-helm-charts/pull/109 because of `imagePullPolicy`, but I have changed how the users secret works.
If `secrets.users` is specified it will not generate the secret, but use the existing one with that name or the operator will generate it if it doesn't exist, but if `secrets.users` is not specified it will generate the secret from the values specified.
I believe this should be now the same as in PXC chart, just need to test if it will make some problem with the upgrade.

PR #101, opened 2021-11-16
K8SPXC-901: Add default resource limits for Operator's pod containers

pooknull:
https://jira.percona.com/browse/K8SPXC-901

Add default resources to `pxc-operator/values.yaml`

### DIFF
--- a/charts/pxc-operator/values.yaml
+++ b/charts/pxc-operator/values.yaml
@@ -34,7 +34,7 @@ resources:
   # lines and add the curly braces after 'resources:'.
   limits:
     cpu: 200m
-    memory: 100Mi
+    memory: 500Mi
   requests:
     cpu: 100m
     memory: 20Mi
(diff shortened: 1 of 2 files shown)


charts/pxc-operator/values.yaml
```diff
@@ -27,17 +27,17 @@ imagePullSecrets: []
...
-  #   cpu: 100m
-  #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  # resources, such as Minikube. If you don't want to specify resources, comment the following
+  # lines and add the curly braces after 'resources:'.
+  limits:
+    cpu: 200m
+    memory: 100Mi
```
tplavcic, line 37: We have increased this to 500Mi because it was leading to OOM on OpenShift environment with 100Mi, so please change it here also.

tplavcic (changes requested):

charts/pxc-operator/values.yaml
pooknull: Resolved in 5a7f4bfaed5157da4420c8e83f91cdf2e1e45c46

PR #108, opened 2021-12-07
Fix source URLs and update the versions for tools

hors:


delgod (approved):

Oneiroi (approved):
LGTM

Oneiroi (approved):
Removal of circleci also fixes the immediate issue, my query would be if circleci is not being used do the scripts used in the other pipeline that is in use require updating also ?

PR #110, opened 2021-12-23
[K8SPSMDB-597] Do no explicitly create namespace when installing PSDMB operator

dalbani:
https://jira.percona.com/projects/K8SPSMDB/issues/K8SPSMDB-597


tplavcic:
We added an annotation for helm to keep the resource and not delete so closing this PR.

PR #122, opened 2022-02-16
Allow mongodb operator to be deployed cluster wide

kvaster:
This PR allows operator to be deployed cluster wide by using `all` as `watchNamespace` value. Also it will create separate `Role` and `RoleBinding` insteadof `ClusterRole` and `ClusterRoleBinding` when deploying with `watchNamespace` when it differs from operator namespace.

Please note that using `watchNamespace` previously wasn't working and will not work even now without patch to operator, cause of a bug: https://github.com/percona/percona-server-mongodb-operator/pull/889

PR #125, opened 2022-03-17
Add support for cluster-wide deployment for mongodb operator

MOZGIII:
An alternative to #122

### DIFF
--- a/charts/psmdb-operator/templates/role-binding.yaml
+++ b/charts/psmdb-operator/templates/role-binding.yaml
@@ -3,7 +3,11 @@ kind: ServiceAccount
 metadata:
   name: {{ include "psmdb-operator.fullname" . }}
 ---
+{{- if or .Values.watchNamespace .Values.watchAllNamespaces }}
+kind: ClusterRoleBinding
+{{- else }}
 kind: RoleBinding
+{{- end }}
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: service-account-{{ include "psmdb-operator.fullname" . }}
@@ -15,11 +19,11 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: {{ include "psmdb-operator.fullname" . }}
-  {{- if .Values.watchNamespace }}
+  {{- if or .Values.watchNamespace .Values.watchAllNamespaces }}
   namespace: {{ .Release.Namespace }}
   {{- end }}
 roleRef:
-  {{- if .Values.watchNamespace }}
+  {{- if or .Values.watchNamespace .Values.watchAllNamespaces }}
   kind: ClusterRole
   {{- else }}
   kind: Role
--- a/charts/psmdb-operator/templates/role.yaml
+++ b/charts/psmdb-operator/templates/role.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.watchNamespace }}
+{{- if or .Values.watchNamespace .Values.watchAllNamespaces }}
 kind: ClusterRole
 {{- else }}
 kind: Role
--- a/charts/psmdb-operator/templates/_helpers.tpl
+++ b/charts/psmdb-operator/templates/_helpers.tpl
@@ -43,3 +43,10 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end -}}
+
+{{/*
+Resolve the effective value of WATCH_NAMESPACE env var.
+*/}}
+{{- define "psmdb-operator.watch-namespace" -}}
+{{- default .Release.Namespace .Values.watchNamespace -}}
+{{- end -}}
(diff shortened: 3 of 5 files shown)


charts/psmdb-operator/templates/role-binding.yaml
```diff
@@ -3,7 +3,11 @@ kind: ServiceAccount
 metadata:
   name: {{ include "psmdb-operator.fullname" . }}
 ---
+{{- if or .Values.watchNamespace .Values.watchAllNamespaces }}
```
jconradhanson, line 6: Should this be `{{ .Values.watchAllNamespaces }}` instead? Or if we wanted to support both values entries for enabling cluster scope, then it'd be `{{- if or not .Values.watchNamespace .Values.watchAllNamespaces }}` to account for negating an empty string given in `watchNamespace`.

We want a `ClusterRoleBinding` for watching all namespaces. So if a `watchNamespace` value is not empty, then we would want a `RoleBinding` to watch the given ns.

charts/psmdb-operator/templates/role-binding.yaml
```diff
@@ -15,11 +19,11 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: {{ include "psmdb-operator.fullname" . }}
-  {{- if .Values.watchNamespace }}
+  {{- if or .Values.watchNamespace .Values.watchAllNamespaces }}
   namespace: {{ .Release.Namespace }}
   {{- end }}
 roleRef:
-  {{- if .Values.watchNamespace }}
+  {{- if or .Values.watchNamespace .Values.watchAllNamespaces }}
```
jconradhanson, line 26: Same comment as above.

charts/psmdb-operator/templates/role.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.watchNamespace }}
+{{- if or .Values.watchNamespace .Values.watchAllNamespaces }}
```
jconradhanson, line 1: Same comment as above.

charts/psmdb-operator/templates/role-binding.yaml
MOZGIII: We either specify that we want to watch all namespaces - we'd need `ClusterRole` to do that, or we specify a single other namespace to watch for (which is a different one from the one we're deployed to, otherwise what's the point) - and we need `ClusterRole` to get access to another namespace too.

We can use `Role` only when we're dealing with the local namespace - and that's where both `watchAllNamespaces` and `watchNamespace` as falsy.

MOZGIII: The `watchNamespace` handling can definitely be improved to handle more edge cases, such as when it's not empty but is explicitly set to a value equal to `.Release.namespace` - but I didn't bother.

Feel free to pick this code up and apply fixes as needed. Tossing it and rewriting this feature from scratch is also an option I'd be fine with.

jconradhanson: Thanks for elaborating. 
I wasn't considering the use-case of deploying the operator to a different namespace than where a single psmdb instance would be deployed. I was making the assumption that if you want RBAC outside of the operator's namespace, then it is more than likely needed cluster-wide. So the way you have it makes sense to me now.

When inspecting the [PXC Operator's helm chart](https://github.com/percona/percona-helm-charts/tree/main/charts/pxc-operator), all three use-cases you outlined are supported (ie. same ns; single, different ns than operator's; cluster-wide) through the `watchNamespace` and `watchAllNamespaces` values. Also, they are using the same if statement structure as what you have, [example here](https://github.com/percona/percona-helm-charts/blob/main/charts/pxc-operator/templates/role-binding.yaml#L11). 

With that said, I'd prefer to have consistency across percona helm charts. So since pxc-operator has a working implementation and is a percona chart, I'd be in favor of adopting the approach there, which is what you have, over what is implemented in https://github.com/percona/percona-helm-charts/pull/122.

jconradhanson:
Commented on the corresponding [roadmap issue here](https://github.com/percona/roadmap/issues/28#issuecomment-1205681142) to seek clarity on when this would be potentially supported.

tplavcic:
@MOZGIII @jconradhanson Cluster wide support for PSMDB operator is scheduled to be released in `1.13.0` release of the operator.
The ticket is located here: https://jira.percona.com/browse/K8SPSMDB-203
(so you can track the progress if you wish, PR is currently in review)

Usually few days after operator release we update the helm charts with all new options, so I expect the same to happen this time.

tplavcic:
Hi @MOZGIII and @conradhanson  !
The PSMDB operator now officially supports cluster-wide mode from version 1.13.0.
Here are the release notes: https://docs.percona.com/percona-operator-for-mongodb/RN/Kubernetes-Operator-for-PSMONGODB-RN1.13.0.html
The change was also implemented in the helm chart in this PR: https://github.com/percona/percona-helm-charts/pull/174
And is available in `psmdb-operator` helm chart version `1.13.1`.
So I will close this PR and if you will find some issues with it or want to provide some other fixes/improvements please open a new ticket/PR.

Thanks!

PR #126, opened 2022-03-25
fix backup.schedule.0.keep to correct default value in readme

stephentreacy:


tplavcic:
Thank you for the fix!

tplavcic (approved):

PR #130, opened 2022-05-03
K8SPSMDB-631 - Update for PSMDB Operator release v1.12.0

tplavcic:


charts/psmdb-db/templates/cluster.yaml
```diff
@@ -15,17 +15,29 @@ spec:
   {{- if .Values.platform }}
   platform: {{ .Values.platform }}
   {{- end }}
-  {{- if .Values.DNSsuffix }}
-  clusterServiceDNSSuffix: {{ .Values.DNSsuffix }}
+  {{- if .Values.clusterServiceDNSSuffix }}
```
tplavcic, line 18: I have renamed this, so that it is consistent with new `clusterServiceDNSMode`, but maybe we would want here `DNSsuffix` and `DNSmode` instead.

charts/psmdb-db/production-values.yaml
```diff
@@ -18,6 +21,9 @@ fullnameOverride: ""
 pause: false
 unmanaged: false
 allowUnsafeConfigurations: false
+multiCluster:
+  enabled: false
+  # DNSSuffix: svc.clusterset.local
```
nmarukovich, line 26: if `enabled: false` we can uncomment `DNSSuffix: svc.clusterset.local` I suppose.

tplavcic: My understanding is that these values are mostly for some sane minimum defaults and some commented examples, so I try to set as less defaults as possible.

egegunes (approved):

cap1984 (approved):

charts/psmdb-db/templates/cluster.yaml
nmarukovich: I believe it will be better to have the value as in CR.

charts/psmdb-operator/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: "1.11.0"
+appVersion: "1.12.0"
 description: A Helm chart for Deploying the Percona Kubernetes Operator for Percona Server for MongoDB
 name: psmdb-operator
 home: https://www.percona.com/doc/kubernetes-operator-for-psmongodb/kubernetes.html
-version: 1.11.0
+version: 1.12.0
 maintainers:
```
nmarukovich, line 7: do we need this 
`- name: paulczar
    email: username.taken@gmail.com`

tplavcic: Thanks, it is updated now!

PR #133, opened 2022-05-05
Add possibility of setting resources for PITR container

hors:


tplavcic (changes requested):
Please bump the chart version.

PR #140, opened 2022-06-10
Update README.md

denisok:
fix set env to pass as string


spron-in (approved):

atymchuk (approved):

tplavcic (changes requested):
Please bump the chart version otherwise chart releaser will fail.

tplavcic (approved):

PR #142, opened 2022-06-11
feat: pmm ingress

dnplkndll:
@denisok not sure why the ingress was not included? 

with the idea to use https internal could remove the ingress 
```
nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
```
by adding 80 to the service
```
ports:
    - port: 80
      targetPort: http
      protocol: TCP
      name: http
```

I tested but to work fine with ingress-nginx and cert-manager.


denisok:
Hi @dnplkndll , thank you for contribution.

No particular reason, just didn't think about ssl termination, all other ingress features are probably useless for pmm so far and it is kinda strange (an maybe there are some side effects) to run ngnix on top of ngnix (that is inside of PMM container).

But ssl termination is quite valid feature. 

@cshiv also suggested ingress, but for the first iteration it was too much and I didn't understand the use case.

Thanks,
Denys

denisok:
@dnplkndll please bump version to 0.2.0

dnplkndll:
not that this is the right place to ask. but
I was kind of curious if there was any consideration to move to more of a sub-chart setup
for example:
https://github.com/sentry-kubernetes/charts/blob/develop/sentry/Chart.yaml

versus the more monolithic image with all the services proxied in the container nginx service. might allow for better resources (requests, limits) config on the specific components?

denisok:
I think I pushed fix for the documentation so please resolve conflicts.

> versus the more monolithic image
Yeah, we are moving towards that but that would take time and not  it would be much optional, maybe ngnix, but other services still would be needed but external.
Currently container is quite tightly coupled.

denisok:
@dnplkndll I plan to review and test it till the end of this week

charts/pmm/templates/ingress.yaml
```diff
@@ -0,0 +1,44 @@
...
+{{- $serviceName := .Values.service.name -}}
+{{- $ingressPathType := .Values.ingress.pathType -}}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "pmm.fullname" . }}
+  labels:
+    {{- include "pmm.labels" . | nindent 4 }}
+  annotations:
+    nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
```
denisok, line 11: to enable ppl to use different ingress controllers, please also parametrize this one. Same as in `helm create` defaul templates:
```
  {{- with .Values.ingress.annotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

or another example here: 
https://github.com/prometheus-community/helm-charts/blob/8f0b863eab3eb604216ce4084608d1b6a7f8e879/charts/kube-prometheus-stack/templates/prometheus/ingress.yaml

charts/pmm/templates/ingress.yaml
```diff
@@ -0,0 +1,44 @@
...
+      http:
+        paths:
+        {{- range .paths }}
+          - path: {{ . }}
+            pathType: {{ $ingressPathType }}
+            backend:
+              service:
+                name: {{ $serviceName }}
+                port: 
+                  number: 443
```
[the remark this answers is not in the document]
dnplkndll: no problem, 80 or 443 or an array? 

I see we have both ports on the app spec.

https://github.com/Kencove/percona-helm-charts/blob/main/charts/pmm/templates/statefulset.yaml#L84

will need mapped in:

https://github.com/percona/percona-helm-charts/blob/main/charts/pmm/templates/service.yaml#L17

```
  ports:
    - port: 443
      targetPort: https
      protocol: TCP
      name: https
```
you could just expose both on the service but otherwise, need 
[] of port and targetPort

https://kubernetes.github.io/ingress-nginx/user-guide/tls/#ssl-passthrough
you can do some type of passthrough. But then seems to be implementation specific. there would need to be some assumption of using the first or a helper needed to pick on > 1.  So I pushed the array with the `use first item` for ingress approach. example is 80 but can leave both, or either one to work. ( don't ask me for the ci test though! ) 

it all my apps we just use http on the app and let the cert-manager manage all the ingress tls. This does work fine with tls as is and with the new options.

charts/pmm/templates/ingress.yaml
dnplkndll: implemented with values example for those needing a hint. (https://github.com/percona/percona-helm-charts/pull/142/commits/86a84ecbdb71b28958a5d325d4037645040c598b)

denisok:
Hi @dnplkndll,
manifest looks ok. Did you test it somehow? I can't make it work with minikube at least, but that might be my setup specific.

What I do:
```
minikube start
minikube addons enable ingress
helm install pmm -f ingress_simple.yaml ./pmm-0.2.0.tgz
```

waiting until ingresa gets ip:
```
$ kubectl get ingress
NAME   CLASS   HOSTS                 ADDRESS        PORTS   AGE
pmm    nginx   chart-example.local   192.168.49.2   80      11m
```

trying to reach by IP, and nothing happens.

```yaml
ingress:
  # ingress.enabled -- Enable ingress controller resource
  enabled: true

  # ingress.hosts -- Ingress resource hostnames and path mappings
  hosts:
    - host: chart-example.local
      paths:
        - /
```

Thanks,
Denys

denisok (approved):

atymchuk (approved):

PR #143, opened 2022-06-15
fix: render pmm-secret from spec

dnplkndll:
So the PMM creds. best practice? I am sure admin is not right but how to you pass from a PMM chart to the db instances from the CRD spec? nothing seems to have that secret rendering now.

not sure if this is a db specific spec. perhaps could be defaulted in the operator? 

@denisok is there a design to default the PMM password? 

is a namespace spec needed other than the chart context?

### DIFF
--- a/charts/pg-db/Chart.yaml
+++ b/charts/pg-db/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: pg-db
 description: 'A Helm chart for Deploying the Percona PostgreSQL database by Percona Distribution for PostgreSQL Operator'
 type: application
-version: 1.2.0
+version: 1.2.1
 appVersion: 1.2.0
 home: https://github.com/percona/percona-postgresql-operator
 maintainers:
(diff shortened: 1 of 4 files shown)


dnplkndll:
> I am not sure how it works exactly in operator. I think `admin` so far is default user that we support and secret comes from operator.
> 
> Recently we started to add ability to authorize by api key, that was added to Percona Server for MySQL.
> 
> But what is the right way to go @cap1984 and @tplavcic should know.

Sounds good, I will bump the chart and await more info. I might find out as we review more. This does work for now. So we can move forward in testing the setup. Thanks

dnplkndll:
I like the token idea. so would you just create that on a pmm user and then pass in the same?

https://github.com/percona/percona-postgresql-operator/search?p=2&q=pmmSecret

I think I understand the goal would be to have the creds, user or token, in the operator spec to pass on. but atm I do not see that, just the name refs and not a template to create it. since the operator does not rely on helm a different approach would surely be needed.

cap1984:
Guys please update README as well

cap1984:
The same approach as it is in pxc and psmdb operators (user/pass) will fit here OK

tplavcic:
I think token support is added to PXC operator currently and if token exists in the secret it will be used, otherwise user/pass will be used like before.

egegunes:
closing this due to inactivity

PR #152, opened 2022-07-20
Add podAnnotations to pxc-operator

pauvos:
Would be nice if we could add some annotations to the pxc-operator pod, e.g. `co.elastic.logs/*`

```sh
helm template test . -f values.yaml --set='podAnnotations.co\.elastic\.logs/json\.message_key=msg'
```

```yaml
spec:
[...]
  template:
    metadata:
      annotations:
        co.elastic.logs/json.message_key: msg
```

### DIFF
--- a/charts/pxc-operator/templates/deployment.yaml
+++ b/charts/pxc-operator/templates/deployment.yaml
@@ -18,6 +18,10 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+    {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
       labels:
         app.kubernetes.io/component: operator
         app.kubernetes.io/name: {{ include "pxc-operator.name" . }}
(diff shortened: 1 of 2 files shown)


charts/pxc-operator/templates/deployment.yaml
```diff
@@ -18,6 +18,10 @@ spec:
     type: RollingUpdate
   template:
     metadata:
+    {{- with .Values.podAnnotations }}
```
tplavcic, line 21: Any reason not to call it just `.Values.annotations`? Looking at our charts only pmm in one place has podAnnotations, all others seem to use just `annotations`

charts/pxc-operator/values.yaml
```diff
@@ -40,3 +40,5 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+podAnnotations: {}
```
tplavcic, line 44: Same as above.

tplavcic (changes requested):
Two more things:
1. Please bump the chart version
2. Update the README.md to add the new option

charts/pxc-operator/templates/deployment.yaml
ebuildy: well podAnnotations is quite standard and a good name for this

tplavcic:
Thank you for the contribution!

PR #153, opened 2022-07-20
update to 2.29.0 and small improvements

denisok:
https://jira.percona.com/browse/PMM-10102

- updated to 2.29.0
- removed init controller (new feature in 2.29.0)
- removed special env - GF_ should be used instead
- improved doc as was suggested during 2.29.0 doc review

### DIFF
--- a/charts/pmm/values.yaml
+++ b/charts/pmm/values.yaml
@@ -14,7 +14,7 @@ image:
   repository: percona/pmm-server
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "2.28.0"
+  tag: "2.29.0"
   imagePullSecrets: []
 
 ## PMM environment variables
@@ -54,7 +54,7 @@ readyProbeConf:
   periodSeconds: 5
   failureThreshold: 6
 
-## PMM secrets
+## @section PMM secrets
 ##
 secret:
   ## @param secret.name Defines the name of the k8s secret that holds passwords and other secrets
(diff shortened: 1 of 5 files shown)


charts/pmm/values.yaml
```diff
@@ -94,46 +95,58 @@ service:
...
-  # ingress.tls -- Ingress TLS configuration
+  ## @param ingress.tls -- Ingress TLS configuration
   tls: []
-  #  - secretName: chart-example-tls
-  #    hosts:
-  #      - chart-example.local
-
+  ##  - secretName: chart-example-tls
+  ##    hosts:
+  ##      - chart-example.local
```
nikita-b, line 147: If we want to have this part as example, I think we need only one `#` here

charts/pmm/values.yaml
```diff
@@ -94,46 +95,58 @@ service:
...
+  ## @param ingress.annotations -- Ingress annotations configuration
   annotations: {}
-    # kubernetes.io/ingress.class: nginx
-    # kubernetes.io/tls-acme: "true"
-    ##  nginx proxy to https
-    # nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
+    ## kubernetes.io/ingress.class: nginx
+    ## kubernetes.io/tls-acme: "true"
+    ###  nginx proxy to https
+    ## nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
```
nikita-b, line 128: If we want to have this part as an example, I think we need only one # here

nikita-b (approved):

tplavcic (approved):

PR #155, opened 2022-07-29
Remove service type ClusterIP, if Ingress enabled

yevhenkizin:
Hello! Therefore, I propose to make such changes for resolve problem, when used AWS Loadbalancer Controller as Ingress Controller:
```
Failed deploy model due to InvalidParameter: 1 validation error(s) found.       │ │ - minimum field value of 1, CreateTargetGroupInput.Port
```
link to issue: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1644

Thank!


denisok:
Hi @yevhenkizin , Thanks for contribution - please update version of the chart, but after #154 to avoid conflict, I will merge that today.

yevhenkizin:
@denisok Sorry, I’ve been busy. Fixed

denisok:
@dnplkndll please check this as well :)

charts/pmm/templates/service.yaml
```diff
@@ -17,7 +14,7 @@ metadata:
     {{- toYaml . | nindent 8 }}
   {{- end }}
 spec:
-  type: {{ $serviceType }}
+  type: {{ $serviceType | default "ClusterIP" }}
```
dnplkndll, line 17: I would remove this as it imply's a default that would not really happen since the service type will always be set.

charts/pmm/templates/service.yaml
```diff
@@ -1,7 +1,4 @@
 {{- $serviceType := .Values.service.type -}}
-{{- if .Values.ingress.enabled -}}
```
dnplkndll, line 2: I have not test a lot of environments. does nginx ingress require the ClusterIP? on digital ocean it was needed. if that were the case, would it would be useful to have a warning in the NOTES.txt, when not both configured ?
```
{{- if .Values.ingress.enabled }}
...
{{- if not contains "ClusterIP" .Values.service.type  }}
nginx ingress may require a service.type of "ClusterIP"
{{- end }}
...
{{- end }}
```

dnplkndll (approved):
these suggestions do not block the overall functionality. just usability suggestions.

charts/pmm/templates/service.yaml
dnplkndll: I tested with NodePort and works fine for the nginx ingress as well. not need to worry about one or the other.

tplavcic (approved):

PR #162, opened 2022-08-11
Avoid re-creating PMM secret every time.

ksa-real:
While in reality the secret is used only the first time, re-creating the
secret creates unnecessary noise for `helm diff`. Also, current
values.yaml design forced user to store there the OIDC password. This is
not desireable especially if values.yaml is persisted in version
control. `secret.create` is kept true, to keep the current behavior by
default.


denisok:
Hi @ksa-real Thank you for contribution!

So you mean that you would have a separate value file for install and for update?

Maybe it is better to use `Release.IsInstall`and just create Secret on Install ?

ksa-real:
> Hi @ksa-real Thank you for contribution!
> 
> So you mean that you would have a separate value file for install and for update?
> 
> Maybe it is better to use `Release.IsInstall`and just create Secret on Install ?

Not really. At this moment the secret is created by the chart. If the secret is not produced in the follow-up chart, then it should be deleted as part of the upgrade. Also, we don't want to store secrets (admin password and OIDC secret) in `values.yaml` which is stored in Git.  (I know there is SOPS but am using a different approach). One way would be retrieving the secret from K8s with lookup(), keep the existing values, and add/generate the not yet existing key/value pairs.

The proposed diff is to keep the current behavior of auto-creating the secret as default and allow manual pre-release secret creation.

denisok:
@ksa-real ok, makes sense. Could you please also update README for this chart with that info (how to store secrets in prod)?

Default option is fine for ppl that just want to try and maybe install everything with default params, so they avoid hassle to create secrets.

Could you also update version, and it good to go.

ksa-real:
Done.

denisok (approved):

tplavcic (approved):

PR #172, opened 2022-09-26
[PSMDB Operator] Fix cluster wide operator mode

dalbani:
In order to get the PSMDB operator to work cluster wide, I had to apply the changes included in the PR.

### DIFF
--- a/charts/psmdb-operator/templates/deployment.yaml
+++ b/charts/psmdb-operator/templates/deployment.yaml
@@ -33,7 +33,7 @@ spec:
           - percona-server-mongodb-operator
           env:
             - name: WATCH_NAMESPACE
-              value: {{ default .Release.Namespace .Values.watchNamespace }}
+              value: {{ if kindIs "invalid" .Values.watchNamespace }}{{ .Release.Namespace }}{{ else }}{{ .Values.watchNamespace | quote }}{{ end }}
             - name: POD_NAME
               valueFrom:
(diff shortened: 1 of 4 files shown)


charts/psmdb-operator/templates/deployment.yaml
```diff
@@ -33,7 +33,7 @@ spec:
           - percona-server-mongodb-operator
           env:
             - name: WATCH_NAMESPACE
-              value: {{ default .Release.Namespace .Values.watchNamespace }}
+              value: {{ if kindIs "invalid" .Values.watchNamespace }}{{ .Release.Namespace }}{{ else }}{{ .Values.watchNamespace | quote }}{{ end }}
```
dalbani, line 36: The trick (hack?) is needed because, contrary the PXC operator chart, no `watchAllNamespaces` property exists.
There was a proposal in https://github.com/percona/percona-helm-charts/pull/125 to use the same structure for the PSMDB operator.

dalbani:
Reading the discussion on https://github.com/percona/percona-helm-charts/pull/125, I saw a point being mentioned that I didn't think about (use case no. 2):
> all three use-cases you outlined are supported (ie. same ns; single, different ns than operator's; cluster-wide)

dalbani:
Good to see the activity in https://github.com/percona/percona-helm-charts/pull/174, so I can close my MR I suppose.

PR #186, opened 2022-12-20
fixed toleration setting pg-db chart

frittenlab:
The current template in the pg-db Chart does not allow the user to specify an array for the tolerations. This is necessary to set the effect and the operator. The current setting only allows me to set a string. Example values.yaml:

```
pgPrimary:
  resources:
    requests:
      memory: "1Gi"
  tolerations:
    - key: node.example.com/storage
      effect: NoSchedule
      operator: Exists
  volumeSpec:
    size: 30G
    storageclass: "local-path"
```

### DIFF
--- a/charts/pg-db/Chart.yaml
+++ b/charts/pg-db/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: pg-db
 description: 'A Helm chart for Deploying the Percona PostgreSQL database by Percona Distribution for PostgreSQL Operator'
 type: application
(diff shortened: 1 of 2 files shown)


egegunes:
@frittenlab could you please fix conflicts and bump version?

frittenlab:
@egegunes I just had a look. The resource pgPrimary does not exist in the current version of the Helm Chart. So my pull request can not be applied. This pull request is over a year old. I think we can close this.

PR #187, opened 2022-12-26
PG operator 2.2.0

cap1984:
Disclaimer: 2.0.0 is a beta version of operators' next gen implementation


tplavcic (approved):
LGTM, but it seems to me that users who don't specify `--version` will install the beta by default and since previously this was installing a GA version I'm not sure that it is optimal. Maybe discuss with @spron-in , although not sure what we can do here if we want to use the same name for the helm chart.

charts/pg-operator/values.yaml
```diff
@@ -1,140 +1,45 @@
...
+image: ""
+
+# set if you want to specify a namespace to watch
+# defaults to `.Release.namespace` if left blank
+# watchNamespace:
+
+# set if operator should be deployed in cluster wide mode. defaults to false
+watchAllNamespaces: false
+
+# set if you want to use a different operator name
```
8000ff, line 18: **issue**: the `operatorName` value doesn't appear to be used by the chart, the comment references `xtradb`
**suggestion**: remove this section from the values file and add comments for the `nameOverride` and `fullnameOverride` fields

charts/pg-db/values.yaml
```diff
@@ -11,6 +11,7 @@ finalizers:
 crVersion: 2.2.0
 image:
   repository: percona/percona-postgresql-operator
+  postgres: ""
```
tplavcic, line 14: Is this a tag or something?
Maybe just call it `tag` like we have in other places.

tplavcic (changes requested):

charts/pg-db/values.yaml
tplavcic: But this is a tag, image consists of repo:tag
The problem currently is that somewhere we have only `image` (like pgbouncer) and somewhere (like for pmm) `image.repo` and `image.tag` and in this case we have `repository` and `image` which is confusing.
So I propose to use the same in all places either just `image` which should really be in format `percona/percona-postgresql-operator:ppg15` (something like that) or `image.repo` and `image.tag` (I think this format we used in other operators).

charts/pg-db/values.yaml
```diff
@@ -1,134 +1,289 @@
...
+#     postgresql:
+#       parameters:
+#         max_parallel_workers: 2
+#         max_worker_processes: 2
+#         shared_buffers: 1GB
+#         work_mem: 2MB
 
 secrets:
-  name:
+ name:
```
tplavcic, line 281: bad indentation?

charts/pg-db/README.md
```diff
@@ -1,135 +1,76 @@
...
 | `defaultUser`                         | Default unprivileged database username                                                    | `pguser`                                   |
 | `defaultDatabase`                         | Default user database created from the cluster start                                                    | `pgdb`                                   |
 | `targetPort`                         | PostgreSQL port                                                    | `5432`                                   |
-| `image.repo`              | PostgreSQL container image repository                                           | `percona/percona-postgresql-operator` |
-| `image.pgver`                     | PostgreSQL container version tag                                       | `ppg14`                              |
+| `repository`              | PostgreSQL container image repository                                           | `percona/percona-postgresql-operator` |
+| `postgresVersion`                     | PostgreSQL container version tag                                       | `14`                              |
```
tplavcic, line 42: seems 15 is default

charts/pg-db/README.md
```diff
@@ -1,135 +1,76 @@
...
-| `pgPrimary.affinity.nodeAffinityType`                     | Node antiAffinity for PostreSQL db pods `preferred/required/disabled` | `preferred`                              |
-| `pgPrimary.affinity.nodeLabel`                     | Key/value map with node lables specified for nodeAffinity | `key/value`                              |
-| `pgPrimary.affinity.advanced`                     | User defined affinity structure to be passed to pods without any processing | `key/value`                              |
 | `pmm.enabled` | Enable integration with [Percona Monitoring and Management software](https://www.percona.com/blog/2020/07/23/using-percona-kubernetes-operators-with-percona-monitoring-and-management/) | `false` |
-| `pmm.image.repository`              | PMM Container image repository                                           | `percona/pmm-client` |
-| `pmm.image.tag`                     | PMM Container image tag                                       | `2.37.0`                              |
+| `pmm.image`              | PMM Container image repository                                           | `percona/pmm-client:2.37.0` |
```
tplavcic, line 48: aren't we missing a lot of new fields like whole `instances` stuff

tplavcic (changes requested):

charts/pg-db/README.md
nmarukovich: it good question - should we put all them in README?

charts/pg-db/values.yaml
nmarukovich: Lint check passed. Any ideas how to check it? 
I checked with 
`ct lint --config .github/ct.yaml --lint-conf .github/lintconf.yaml`
it was successful, with 2 spaces check failed.

charts/pg-db/README.md
tplavcic: yes, we need to document each field which can be changed in the values

charts/pg-db/values.yaml
tplavcic: I think it's confused because of inconsistency, when you specify 2 spaces everywhere the lint is ok. I just used some online yaml formatter and it formatted the whole yaml with 2 spaces and it passes the lint.

charts/pg-db/templates/cluster.yaml
```diff
@@ -13,173 +13,81 @@ metadata:
...
-      {{- if .Values.pgPrimary.affinity.nodeLabel }}
-      nodeLabel:
-{{ toYaml .Values.pgPrimary.affinity.nodeLabel | indent 8 }}
-      {{- end }}
-      {{- if .Values.pgPrimary.affinity.advanced }}
-      advanced:
-{{ toYaml .Values.pgPrimary.affinity.advanced | indent 8 }}
+  instances:
+    - name: instance1
+      replicas: 3
```
tplavcic, line 36: We cannot have hardcoded values in the template file.

charts/pg-db/templates/cluster.yaml
```diff
@@ -13,173 +13,81 @@ metadata:
...
+      dataVolumeClaimSpec:
+        accessModes:
+        - ReadWriteOnce
+        resources:
+          requests:
+            storage: 5Gi
```
tplavcic, line 42: Hardcoded.

charts/pg-db/templates/cluster.yaml
```diff
@@ -13,173 +13,81 @@ metadata:
...
+    pgbackrest:
+      image: {{ include "pg-database.backup-image" . }}
+      manual:
+        repoName: repo1
+        options:
+         - --type=full
+      repos:
+      - name: repo1
+        schedules:
+          full: "0 0 * * 6"
```
tplavcic, line 86: Hardcoded.

charts/pg-db/templates/cluster.yaml
```diff
@@ -13,173 +13,81 @@ metadata:
...
+        volume:
+          volumeClaimSpec:
+            accessModes:
+            - ReadWriteOnce
+            resources:
+              requests:
+                storage: 1Gi
```
tplavcic, line 93: Hardcoded.

charts/pg-db/templates/cluster.yaml
```diff
@@ -13,173 +13,81 @@ metadata:
...
+          memory: {{ .Values.proxy.pgBouncer.resources.requests.memory }}
+        {{- end}}
+        {{- if .Values.proxy.pgBouncer.resources.limits }}
+        limits:
+          cpu: {{ .Values.proxy.pgBouncer.resources.limits.cpu }}
+          memory: {{ .Values.proxy.pgBouncer.resources.limits.memory }}
+        {{- end}}
+      {{- end}}
+      {{- if .Values.proxy.pgBouncer.expose }}
+      expose:
```
tplavcic, line 62: We seem to be missing fields which we have in values like: proxy.pgbouncer..sidecars, proxy.pgbouncer.tolerations, proxy.pgbouncer.topologySpreadConstraints, proxy.pgbouncer.config, proxy.pgbouncer.customTLSSecret and some others probably.

tplavcic (changes requested):

tplavcic (approved):

hors (approved):

PR #188, opened 2023-01-03
K8SPS-188 - Update ps-operator and ps-db charts for 0.4.0 release

tplavcic:


cap1984 (approved):

nmarukovich (approved):

nmarukovich (approved):

charts/ps-operator/templates/role-binding.yaml
```diff
@@ -44,3 +52,25 @@ roleRef:
...
+  apiGroup: rbac.authorization.k8s.io
+  {{- if .Values.watchNamespace }}
+  kind: ClusterRole
+  {{- else }}
+  kind: Role
+  {{- end }}
+  name: {{ include "ps-operator.fullname" . }}-orchestrator
+subjects:
+- kind: ServiceAccount
+  name: {{ include "ps-operator.fullname" . }}-orchestrator
```
cap1984, line 75: Is it ok to keep auto generated service account name reference here when `serviceAccount.create=false`?

cap1984 (approved):

charts/ps-operator/templates/role-binding.yaml
tplavcic: You are right, it might be problematic, but we have a general issue with orchestrator service account so we will need changes here later anyway, I have opened a ticket here: https://jira.percona.com/browse/K8SPS-235

nmarukovich (approved):

PR #190, opened 2023-01-09
CLOUD-758 - Add option to skip SAs and RBAC

iblackman:
Adds the option to skip ServiceAccounts and RBAC. Similar to how it is done in https://github.com/percona/percona-helm-charts/tree/main/charts/pg-operator

### DIFF
--- a/charts/pxc-operator/Chart.yaml
+++ b/charts/pxc-operator/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: 1.12.0
+appVersion: 1.12.1
 description: A Helm chart for deploying the Percona Operator for MySQL (based on Percona XtraDB Cluster)
 name: pxc-operator
 home: https://docs.percona.com/percona-operator-for-mysql/pxc/
-version: 1.12.0
+version: 1.12.1
 maintainers:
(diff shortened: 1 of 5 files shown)


iblackman:
Hi @gen1us2k @denisok Could any of you review this PR? Thanks in advance.

denisok (approved):

cap1984:
Hey @iblackman 
Could you please elaborate more on intended use-case? PXC and PG operator manage differently the rbac settings

iblackman:
> Hey @iblackman Could you please elaborate more on intended use-case? PXC and PG operator manage differently the rbac settings

Sure. In our use case we have different workflows that deploy resources where our normal CD doesn't have permission to deploy RBAC resources as a safety measure, so we normally extract that to a different workflow. The change I suggest contemplate the case (can skip RBAC) and doesn't change anything to the normal behavior intended with the chart prior my change.

denisok (approved):

tplavcic (approved):

tplavcic:
Hi @iblackman ! Thank you very much for your contribution! It has now been included for pxc-operator chart and we will likely include it in other operator charts as well since we look that they have same/similar options.

PR #194, opened 2023-01-23
K8SPXC-1188 - pxc-db - persist data by default, use random passwords

tplavcic:
Changes:
- by default use random passwords generated by operator
- remove production values and merge them to default values file
- enable persistence by default
- put annotation `resource-policy: keep` on cluster namespace when operator is set to watch namespace (this namespace.yaml will be removed sometime later but for now we will use this annotation)
- disable finalizers `delete-pxc-pvc` and `delete-proxysql-pvc` by default


spron-in (approved):

charts/pxc-operator/Chart.yaml
```diff
@@ -3,10 +3,8 @@ appVersion: 1.12.0
 description: A Helm chart for deploying the Percona Operator for MySQL (based on Percona XtraDB Cluster)
 name: pxc-operator
 home: https://docs.percona.com/percona-operator-for-mysql/pxc/
-version: 1.12.1
+version: 1.12.2
 maintainers:
```
nmarukovich, line 7: could we use 1.12.3 too?

PR #199, opened 2023-02-21
default do not delete pvc, as per the documentation 

titansmc:
default do not delete pvc, as per the documentation https://docs.percona.com/percona-operator-for-mongodb/operator.html?h=finalize


tplavcic:
Hi @titansmc !
Thank you for contribution. We had a release of PSMDB operator 1.14 together with the charts and this option was changed there.
I'm sorry I didn't catch that you created a PR about the same thing otherwise I would merge it to release branch, but still the change which you proposed is now applied, but in 1.14 version.

PR #200, opened 2023-02-24
Fix ingress grpc

johnwc:
This PR fixes the issue with gRPC that occurs when using an ingress. I fully tested it using ingress controller from NGINX Inc. I have not been able to test it using the ingress from Kubernetes official, but the expected changes should be in place for it to work based on their documentation.


charts/pmm/templates/service.yaml
```diff
@@ -21,3 +21,29 @@ spec:
   {{- end }}
   selector:
     {{- include "pmm.selectorLabels" . | nindent 4 }}
+{{- if .Values.ingress.enabled }}
```
denisok, line 24: should it be `{{- if $.Values.ingress.nginxInc }}` ?

denisok:
Hi @johnwc,

Thank you very much for your contribution!

Could you please also update README ?

I am using https://github.com/bitnami-labs/readme-generator-for-helm to update tables, but you could do it manually.

Also you need to bump chart version to pass tests.

Thanks,
Denys

charts/pmm/templates/service.yaml
johnwc: No, the service is only valid if ingress is enabled. Does not matter the ingress type.

denisok (approved):

denisok:
@dnplkndll could you please review as well ?

denisok:
@tplavcic @gen1us2k please review

gen1us2k (approved):

gen1us2k:
Hi @johnwc,

Thanks for your contribution! Looks good to me

Aleksa-del:
Hello, @johnwc ! Thank you for your contributions! We would like to send you a gift from Percona. Please, contact us at community-team@percona.com.

PR #204, opened 2023-03-08
Removing redundant pxc-db crd files

Jancis:
We're using namespace limited SA for deployment and since `pxc-db` also wants to install crds, our deployment fails. `pxc-db` crd is identical to `pxc-operator` crd, it's already set up, since having an operator is a [pre-requisite](https://github.com/percona/percona-helm-charts/tree/main/charts/pxc-db#pre-requisites).


tplavcic:
@Jancis FYI: We have merged the change for `pxc-db`, `psmdb-db` and `ps-db` charts.

PR #220, opened 2023-05-12
Add helm chart prom-rds-exporter

dungpham91:
Hi,

- I created a Helm chart for rds_exporter and named it prom-rds-exporter.
- I tested in my EKS cluster and it's working fine.
- Also created a README for this chart, by it, you can install it easily.

If you guys can take few mins to review and approve it. I think it's will be more convenience for the other guys when using it.

Thanks.

PS: my old PR https://github.com/percona/rds_exporter/pull/89


jagadeeshmedavarapu1:
Can we define cross-account role for RDS-Exporter..? Is it supportable?

dungpham91:
> Can we define cross-account role for RDS-Exporter..? Is it supportable?

I only have one account so I've never tried this. You can try it with cross-account role. In theory, the role for the exporter only needs to ensure the policies as in the README I wrote.

BupycHuk (dismissed):
Looks good from rds_exporter configuration side. @tplavcic can someone from Cloud Native team review from K8s perspective?

tplavcic (dismissed):

charts/prom-rds-exporter/values.yaml
```diff
@@ -0,0 +1,233 @@
...
+    - host: chart-example.local
+      paths:
+        - path: /
+          pathType: ImplementationSpecific
+  tls: []
+  #  - secretName: chart-example-tls
+  #    hosts:
+  #      - chart-example.local
+
+securityContext:
```
egegunes, line 222: securityContext is duplicated

dungpham91: Hi @egegunes 
Thanks for comment. I've updated the chart and also check `helm lint` from my locally. Please verify again. Thanks.

PR #227, opened 2023-07-09
K8SPG-402 Remove duplicate "pause" key from cluster.yaml

e3b0c442:
Helm install fails with pg-db 2.2.0 due to the existence of a duplicate "pause" key in the cluster.yaml template This PR removes the duplicate.

### DIFF
--- a/charts/pg-db/templates/cluster.yaml
+++ b/charts/pg-db/templates/cluster.yaml
@@ -20,7 +20,6 @@ spec:
   postgresVersion: {{ .Values.postgresVersion}}
   standby:
     enabled: {{ .Values.standby.enabled }}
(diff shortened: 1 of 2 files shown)


charts/pg-db/templates/cluster.yaml
```diff
@@ -20,7 +20,6 @@ spec:
   postgresVersion: {{ .Values.postgresVersion}}
   standby:
     enabled: {{ .Values.standby.enabled }}
-  pause: {{ .Values.pause }}
```
hors, line 23: Please bump version of the chart.

tplavcic:
@e3b0c442 Thank you very much for your contribution!

PR #231, opened 2023-07-16
Update cluster.yaml

ssk199441:
Following modules are fixed

user creation
database creation
database init
patroni config


hors:
@ssk199441 could you please fix the conflicts and bump the chart version? Thank you.

PR #239, opened 2023-08-21
K8SPXC-1291 - Add automatic /tmp mount when readOnlyRootFileSystem is set

MarcThe:
When using the container security context _readOnlyRootFileSystem_ the container cannot start because it cannot write to _/tmp_.

This adds an automatic _/tmp_ mount when _readOnlyRootFileSystem_ security context is set.

### DIFF
--- a/charts/pxc-operator/templates/deployment.yaml
+++ b/charts/pxc-operator/templates/deployment.yaml
@@ -88,7 +88,7 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
     {{- end }}
-    {{- with .Values.containerSecurityContext.readOnlyRootFileSystem }}
+    {{- if .Values.containerSecurityContext.readOnlyRootFilesystem }}
       volumes:
         - name: tmpdir
           emptyDir: {}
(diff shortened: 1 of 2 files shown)


charts/pxc-operator/templates/deployment.yaml
```diff
@@ -40,6 +40,11 @@ spec:
             protocol: TCP
           command:
           - percona-xtradb-cluster-operator
+          {{- with .Values.containerSecurityContext.readOnlyRootFileSystem }}
```
tplavcic, line 43:
```suggestion
          {{- if .Values.containerSecurityContext.readOnlyRootFilesystem }}
```

charts/pxc-operator/templates/deployment.yaml
```diff
@@ -83,6 +88,11 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
     {{- end }}
+    {{- with .Values.containerSecurityContext.readOnlyRootFileSystem }}
```
tplavcic, line 91:
```suggestion
    {{- if .Values.containerSecurityContext.readOnlyRootFilesystem }}
```

tplavcic (changes requested):

tplavcic (approved):

tplavcic:
@MarcThe Thank you for the contribution!

PR #242, opened 2023-09-29
K8SPSMDB-997 - Add serviceAccountName to the existing resources

Rockawear:
This is to address issues encountered in the link below. Basically, PSMDBBackup aren't working when using `serviceAccountName` because the entire cluster is using `default`  when using the psmdb-db chart. Since the crd supports serviceAccoountName already, this change adds the ability to set `serviceAccountName` as needed. Ideally, we would want just the specific container to have this kind of access. After some research, it doesn't seem possible to container level service account.

https://forums.percona.com/t/how-to-setup-aws-s3-storage-backup-using-service-account-irsa/21672


tplavcic:
@Rockawear Thank you very much for the contribution! Your commits were merged and are available in chart version 1.15.0.

PR #244, opened 2023-10-04
CLOUD-838 - Fix namespace assignment in pxc-db s3-secret.yaml

artemdanielyan:
Fix nil pointer evaluating interface {}.Namespace, as in this case Namespace is used in "range" helm namespace


egegunes:
@artemdanielyan please bump the version

tplavcic (approved):

tplavcic:
Thank you very much for the nice fix!

nmarukovich (approved):

PR #252, opened 2023-10-12
CLOUD-838 - Fix dataVolumeClaimSpec...storage in README

mrrsm:
`instances.walVolumeClaimSpec.resources.requests.storage` was duplicated for dataVolumeClaimSpec


egegunes:
@mrrsm please bump the version

spron-in (dismissed):

mrrsm:
@egegunes Do you want a version bump for a docs only change?

egegunes:
@tplavcic does the requirement for bump caused by the lint-test check or does helm require bump for README change?

tplavcic:
@egegunes @mrrsm Each change in helm charts triggers release so the version needs to be bumped since we cannot re-release old version. I agree it's a bit silly in this case, I'll check if chart releaser and linter can be configured differently, but for now it is like this.

tplavcic (approved):

tplavcic:

PR #261, opened 2023-11-21
Update CODEOWNERS for pmm helm

rnovikovP:
Let's update it according to current roles

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-/charts/pmm/ @gen1us2k @atymchuk @denisok @spron-in
+/charts/pmm/ @BupycHuk @ademidoff @spron-in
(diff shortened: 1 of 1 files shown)


ademidoff (dismissed):

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-/charts/pmm/ @gen1us2k @atymchuk @denisok @spron-in
+/charts/pmm/ @BupycHuk @ademidoff @spron-in
```
BupycHuk, line 1: let's use teams instead of exact persons. percona/pmm-review-be should be the best option here

rnovikovP, line 1:
```suggestion
/charts/pmm/ @percona/pmm-review-be  @spron-in
```

spron-in (approved):

BupycHuk (approved):

PR #173, opened 2022-09-26
CLOUD-838 - [PSMDB Operator] Support health and metrics endpoints

dalbani:
Or is there a reason I'm not aware of why health and metrics endpoints are not used?

PR #265, opened 2023-11-27
Change container port to match metrics-bind-address

kovaxur:
The 'metrics' container port is now set to 60000, but the default value for the operator's metric endpoint is 8080:

```
  -metrics-bind-address string
        The address the metric endpoint binds to. (default ":8080")
```

Without changing to 8080, it is impossible to monitor the operator via Prometheus.

This could be further refined to allow the user to set a custom port for monitoring, but I think this is not important now.


hors:
@kovaxur please sign CLA

kovaxur:
@hors done, thanks

egegunes:
@kovaxur thank you for your contribution. there is another PR (#173) which implements the same things and additionally configures probes so we'll merge than one. I'm closing this one.

PR #270, opened 2023-12-06
CLOUD-838 - psmdb-db fix: update path and type HostPath

qmiinh:
The setting [HostPath](https://docs.percona.com/percona-operator-for-mongodb/storage.html#hostpath) is not working.
Here is the helm generator if config the volumeSpec.hostPath.path:
```
    volumeSpec:
      hostPath:
        path: map[path:<value>]
        type: Directory
```


spron-in:
@egegunes do we need to bump the version?

egegunes:
> @egegunes do we need to bump the version?

yes

tplavcic (approved):

tplavcic:

Issue #292, opened 2024-02-20
Make namespace creation as optional 


sumeet-zuora:
Can we make the creation of namespace as optional at https://github.com/percona/percona-helm-charts/blob/main/charts/psmdb-operator/templates/namespace.yaml as we might already have namespace created from different automation

spron-in:
Hello @sumeet-zuora ,

yeah, we have it in our plans and it might be delivered under https://perconadev.atlassian.net/browse/K8SPSMDB-930

sumeet-zuora:
Can this be also added to pxc, pg and ps operator seems all have same issue

tplavcic:
@sumeet-zuora From what I see pxc is affected and I will create a PR for it, ps and pg don't seem to be affected since they don't create namespaces (but correct me if I'm wrong and don't see something).

abh:
in `ps-operator` it'd be nice if it'd be possible to set the `WATCH_NAMESPACE` variable through the chart, instead of having it fixed to the namespace of the operator.

PR #296, opened 2024-03-05
K8SPSMDB-930 - psmdb-operator - Fix watchNamespace

tplavcic:
If we specify `watchNamespace` with multiple namespaces separated with commas namespace creation will fail (because it will try to create namespace `aaa, bbb`) so now we do it in a loop for each specified namespace and only if `createNamespace` is set to `true`.

**Default is now being changed to not create a namespace unless enabled.**


nmarukovich (approved):

spron-in (approved):

spron-in:
fixes #292

PR #300, opened 2024-03-12
K8SPXC-1365 - pxc-operator - Fix watchNamespace

tplavcic:
If we specify `watchNamespace` with multiple namespaces separated with commas namespace creation will fail (because it will try to create namespace `aaa, bbb`) so now we do it in a loop for each specified namespace and only if `createNamespace` is set to `true`.

**Default is now being changed to not create a namespace unless enabled.**


charts/pxc-operator/README.md
```diff
@@ -38,6 +38,8 @@ The chart can be customized using the following configurable parameters:
 | `logStructured`                 | Force PXC operator to print JSON-wrapped log messages                                          | `false`                                          |
 | `logLevel`                      | PXC Operator logging level                                                                     | `INFO`                                           |
 | `disableTelemetry`              | Disable sending PXC Operator telemetry data to Percona                                         | `false`                                          |
+| `watchNamespace`                | Set when a different from default namespace is needed to watch (comma separated if multiple needed)  | `""`               |
```
ptankov, line 41: I would like to suggest a different wording here:

Comma separated list of namespace(s) to watch when different from "default"

charts/pxc-operator/README.md
```diff
@@ -38,6 +38,8 @@ The chart can be customized using the following configurable parameters:
...
+| `createNamespace`               | Set if you want to create watched namespaces with helm                                         | `false`                                   |
```
ptankov, line 42: I would like to suggest a different wording here:

Create the watched namespace(s)

ptankov (approved):

ptankov (approved):

PR #310, opened 2024-03-21
K8SPG-529  - Fix lint issues and un-wrapped template values

justicel:
There are several variable/template values that are un-wrapped, which cause compilation issues in helm, for pg-db. This fixes those, and also a few lint issues.


egegunes:
@justicel please bump up the chart version

tplavcic:
@justicel Please bump the version one more time (there were some other merges so it is needed) and it seems I don't have a permission to make a commit on your branch.

tplavcic:
@justicel Thank you very much for the contribution! It was merged now.

Issue #305, opened 2024-03-17
pg-db - Syntax issue while helm install 


TAbhay:
While setting up secret in backups, it's not able to detect the secret 'pg-backup-secret' while helm install.

Issue:
![issue](https://github.com/percona/percona-helm-charts/assets/56065895/73e08f1c-b99f-4c59-b0bb-7103acfcd9ab)

Fix:
Syntax needs to be changed in values.yaml
![fix-](https://github.com/percona/percona-helm-charts/assets/56065895/9f8c61df-c00c-428e-a840-515b57d74694)

TAbhay:
Raised Pull request : https://github.com/percona/percona-helm-charts/pull/313
@tplavcic  Please review it. 
Thank you. 🙂

tplavcic:
The PR is merged (with some changes) so this should be fixed now.

PR #313, opened 2024-03-26
K8SPG-540  - pg-db -  Syntax issue while helm install

TAbhay:
https://github.com/percona/percona-helm-charts/issues/305#issue-2190589498

While setting up secret in backups, it's not able to detect the secret 'pg-backup-secret' while helm install.

Issue:
issue

Fix:
Syntax needs to be changed in values.yaml
fix-


tplavcic (approved):

tplavcic:
Thank you for the contribution! I have changed your PR to make the template have the same syntax as in original custom resource. It is now merged.

PR #318, opened 2024-04-08
K8SPG-546  - pg-db: restore without postgresCluster, options for pgbackrest, restore from gcs and azure

Ratio2:


tplavcic (approved):

tplavcic:
@Ratio2 Thank you very much for the nice contribution! It is now merged.

PR #324, opened 2024-04-29
K8SPG-572 - Fix indentation issue with pgbouncer config

justicel:
If you use pgbouncer configs with the chart, it doesn't render properly. This fixes that.


tplavcic (approved):

tplavcic:
@justicel Thank you very much for the fix! It was merged now.

PR #330, opened 2024-06-03
[pmm] add extraObjects

TheRealNoob:
add extraObjects

extraObjects can be defined as strings or dictionaries.  string format is useful if the field names are variable, for example...
```
apiVersion: v1
kind: Secret
type: Opaque
data:
  {{- range $key, $value := .Values.someDict }}
    {{ $key }} : {{ b64enc $value }}
  {{- end }}
```


skhtor:
Also eager for this to be merged. YAML merging is very useful and keeps our env-specific values files small.

PR #331, opened 2024-06-04
add `annotations` value

alex-souslik-hs:
Add `annotations` value to allow custom annotations on the `PerconaServerMongoDB` resource.

### DIFF
--- a/charts/psmdb-db/Chart.yaml
+++ b/charts/psmdb-db/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.16.0"
 description: A Helm chart for installing Percona Server MongoDB Cluster Databases using the PSMDB Operator.
 name: psmdb-db
 home: https://www.percona.com/doc/kubernetes-operator-for-psmongodb/index.html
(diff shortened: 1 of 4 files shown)


cshiv:
The ability to modify is always good.LGTM.
@tplavcic Do you see any added value in having the below default annotation? Can it be removed?
```
  kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"psmdb.percona.com/v1","kind":"PerconaServerMongoDB"}
```

alex-souslik-hs:
@cshiv @tplavcic if it can be removed entirely that'd be event better, it causes issues in Argo CD for example.

tplavcic:
@alex-souslik-hs I cannot recall why this custom annotation was added. Please remove it in your PR and we will try to re-check it before merging your PR.
Thanks for contribution!

alex-souslik-hs:
@tplavcic any update?

tplavcic:
@alex-souslik-hs The update is that it is merged! ;) Thank you very much for the contribution!

PR #350, opened 2024-07-15
Default network Policy and extraObjects

jehutyy:
The actual chart overlook network policy thus leading to mongo database not being able to reconcile within an environment that uses network policies.

I introduce two netpol : one which allows intra-namespace communication and another one that allows communication between db namespace and operator namespace.

Additionally, there's an extraObjects value that can be used when a use case isn't covered by available values in values file.


gkech:
Isn't that something a user of the operator can do and not have it integrated on the helm chart of the operator? Why do we need to have it here? If this is the case, we can close this PR.

PR #366, opened 2024-09-02
K8SPG-588: Add trackLatestRestorableTime

egegunes:


hors:
@egegunes I think we can't use main :( we need to have some release branch  @jvpasinatto WDYT?

jvpasinatto:
@egegunes @hors Yes, please commit in [K8SPG-604-pg-250](https://github.com/percona/percona-helm-charts/tree/K8SPG-604-pg-250) branch, so we can try to bump the chart version only once to 2.5.0

charts/pg-db/values.yaml
```diff
@@ -311,6 +311,7 @@ proxy:
 #          pool_mode: transaction
 
 backups:
+  trackLatestRestorableTime: true
```
jvpasinatto, line 314: Please add this parameter in the readme

jvpasinatto (changes requested):

charts/pg-db/values.yaml
egegunes: https://github.com/percona/percona-helm-charts/pull/366/commits/7dd7151278231e4499304e488f95e51f3e0b1cad

jvpasinatto (approved):

Issue #333, opened 2024-06-05
pg-db chart does not support tolerations as list


networkhermit:
chart version: 2.3.18

There are multiple examples that demonstrate the usage of tolerations in the values file:

https://github.com/percona/percona-helm-charts/blob/4d243856a0e272510a74d3c696b67c13d0702a5a/charts/pg-db/values.yaml#L118-L122

But the helm template does not support them well, passing a tolerations list when installing a helm release would fail:

https://github.com/percona/percona-helm-charts/blob/4d243856a0e272510a74d3c696b67c13d0702a5a/charts/pg-db/templates/cluster.yaml#L169-L175

Current workaround is passing a single tolerations map:

```yaml
  tolerations: 
    effect: NoSchedule 
    key: role 
    operator: Equal 
    value: connection-poolers
```

jvpasinatto:
Fixed in #369

PR #369, opened 2024-09-04
K8SPG-644 - Fix tolerations iteration issue for pg-db cluster template

kv-dhyey-moliya:
Currently the cluster template incorrectly takes only single value (object) for tolerations instead of accepting an array of objects. 

There is already an issue for this : https://github.com/percona/percona-helm-charts/issues/333
This PR fixes that issue by updating the template code that Instead of replacing with single value, replaces array of tolerations for pg-db cluster template.

### DIFF
--- a/charts/pg-db/templates/cluster.yaml
+++ b/charts/pg-db/templates/cluster.yaml
@@ -222,10 +222,10 @@ spec:
       {{- if $instance.tolerations }}
       tolerations:
       {{- range $toleration :=  $instance.tolerations }}
-      - effect: {{ $toleration.effect }}
-        key: {{ $toleration.key }}
-        operator: {{ $toleration.operator }}
-        value: {{ $toleration.value }}
+        - effect: {{ $toleration.effect }}
+          key: {{ $toleration.key }}
+          operator: {{ $toleration.operator }}
+          value: {{ $toleration.value }}
       {{- end }}
       {{- end }}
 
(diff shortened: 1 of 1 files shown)


tplavcic (changes requested):
Thank you for the contribution! Please bump the chart version.

kv-dhyey-moliya:
This fix might break the existing setup of users as the workaround was to put single `map` under tolerations, and it would work. But the fix makes the chart use `range`. 

What should be ideal version bump in this case? Do the charts use semantic versioning? Then it might require to increase major version due to incompatibility with older versions. Changing major version seems like an overkill.

Or else we can resort to conditional logic if the user passes array, use the array, else use the direct map values as shown below.
```
tolerations:
{{- $tolerations := .Values.mapItems }}
{{- if eq (kindOf $tolerations) "map" }}
  - key1: {{ $tolerations.key1 }}
    key2: {{ $tolerations.key2 }}
{{- else}}
{{- range .Values.mapItems }}
  - key1: {{ .key1 }}
    key2: {{ .key2 }}
{{- end }}
{{- end}}
```

Please suggest.

tplavcic:
@kv-dhyey-moliya Usually we bump the "patch" version, but if it's something that breaks compatibility then we merge it on the next operator release when the "minor" version is bumped. At the current moment we do not bump "major" version at all. Basically our helm chart versions follow the operator version except that we can bump the "patch" version if there's a small fix just for the helm chart.
So I propose for your change that we merge it on the next PG operator release.

kv-dhyey-moliya:
@tplavcic thank you for the update. Then, should I update the chart version to a major version for this PR? Or leave it as it is for now?

Additionally, As per my other suggestion, we can make it backwards compatible if we do conditional check, then we can update with a patch version, but again that will add unnecessary conditional code into the codebase. Ping me if you feel it's ok for now to add the conditional code, then I'll update it accordingly with a patch version.

tplavcic:
I think just leave it for now, when we will work on the next release we will merge your fix to the release branch and make any additional changes needed.

jvpasinatto (approved):

PR #377, opened 2024-09-19
K8SPSMDB-1160: disable pvc resizing by default

pooknull:
https://perconadev.atlassian.net/browse/K8SPSMDB-1160

### DIFF
--- a/charts/psmdb-db/values.yaml
+++ b/charts/psmdb-db/values.yaml
@@ -28,6 +28,8 @@ unsafeFlags:
   terminationGracePeriod: false
   backupIfUnhealthy: false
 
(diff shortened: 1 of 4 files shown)


charts/psmdb-db/values.yaml
```diff
@@ -28,7 +28,7 @@ unsafeFlags:
   terminationGracePeriod: false
   backupIfUnhealthy: false
 
-expansionEnabled: false
+enableVolumeExpansion: false
```
jvpasinatto, line 31: Please, add this new field in readme

jvpasinatto (changes requested):

PR #382, opened 2024-09-21
add the option to load env from a external secet

KyriosGN0:
the reasom is that today, when using an existing secret, we can't put different values from the ones that are in the created secret. 
this will allow users to use an external secret (for example with external-secrets operator) and put any required variables as env vars


spron-in:
I like it. @BupycHuk what do you think?

BupycHuk (dismissed):
Looks good to me, please fix conflicts and update readme

KyriosGN0:
@BupycHuk fixed conflicts, updated readme (although im not sure where should i have updated it)

KyriosGN0:
@BupycHuk possibly any updates here?

BupycHuk (approved):

BupycHuk (approved):

PR #383, opened 2024-09-25
EVEREST-1511 | Everest Helm chart

mayankshah1607:
Adds a new Helm chart for Everest.

TODO:
- Chart release automation. For now it has been skipped from releases since we need a different release strategy than what is provided in this repo.
- Chart has been excluded from tests, since it cannot be installed on any random namespace. Will be fixed as a part of the previous point.
- Add docs for ArgoCD / Flux integration.

Since the chart is not yet published, you can test it locally:

1. Provision Everest system components:
```sh
helm dependency update charts/everest
helm dependency update charts/everest/charts/everest-db-namespace
helm install everest-core ./charts/everest --namespace everest-system --create-namespace
```

2. Provision a DB namespace:
```sh
cat <<EOF | kubectl create -f -
apiVersion: v1
kind: Namespace
metadata:
  name: everest
  labels:
    app.kubernetes.io/managed-by: everest
EOF
helm install everest-db ./charts/everest/charts/everest-db-namespace --namespace everest
```
3. Uninstallation
```sh
helm uninstall everest-db -n everest
kubectl delete namespace everest
helm uninstall everest-core -n everest-system
kubectl delete namespace everest-system
```
Important points to know:
- Due to several technical limitations, this chart doesn't work correctly with GitOps tools like ArgoCD or Flux. There are workarounds which will be added as separate docs later.
- Do not try to install anything other than dev-latest (0.0.0) otherwise it will not work either.


.github/CODEOWNERS
```diff
@@ -1,2 +1,3 @@
 * @hors @tplavcic @nmarukovich @ptankov @jvpasinatto @eleo007
 /charts/pmm/ @percona/pmm-review-be  @spron-in
+/charts/everest/ @percona/everest-backend
```
mayankshah1607, line 3: @recharte it says there's an error here, possibly related to permissions. Do you know what's wrong? I was under the impression all of us under this team had write access to this repo 🤔

.github/workflows/test.yaml
```diff
@@ -48,7 +49,7 @@ jobs:
           if [ -n "$OPERATOR_INSTALL" ]; then
             for operator in $OPERATOR_INSTALL; do helm install --namespace default $operator charts/$operator/.; done
           fi
-          ct install --config .github/ct.yaml
+          ct install --config .github/ct.yaml --excluded-charts=everest
```
mayankshah1607, line 52: Had to exclude it for now because this tool tries to deploy in a random namespace, but the chart fails if it is anything other than `everest-system`

At the same time, I could not figure out how to set a namespace only for specific charts.

mayankshah1607: Maybe there can be a separate workflow for everest, since we'll need one anyway for running a manual release

charts/everest/templates/everest-olm/packageserver.apiservice.yaml
```diff
@@ -0,0 +1,35 @@
+{{- $tls := fromYaml (include "olm.certs" .) }}
+---
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  name: v1.packages.operators.coreos.com
```
mayankshah1607, line 6: This chart changes the way OLM is deployed. Earlier we applied a CSV for the packageserver component, but this lead to many issues. For example, when running `helm uninstall ...` the OLM operator always got deleted first, and the CSV finalizer was stuck. As a result, the entire `everest-olm` namespace got stuck.

By deploying the components manually (using native objects such Deployment, APIService, etc.) helm handles deletion correctly..

.github/CODEOWNERS
recharte, line 3:
```suggestion
/charts/everest/ @percona/everest-backend-devs
```

recharte (approved):
I didn't go through the 11.9k lines of yaml in detail as it's pretty much impossible but I did run quite a few tests and only found one issue about the `everest-olm` and `everest-system` showing up with DBEngines, this appears to be related with https://github.com/percona/everest-operator/pull/534 so I'm approving this PR.

PR #391, opened 2024-10-16
github issue 337, added resources block under pmm

siddharthmakwana:
GitHub issue 337, added resources block under pmm to limit resources like CPU and mem and this also helps maintain the QoS of pods. Currently, due to a lack of resources block under PMM, k8s won't ensure Guaranteed QoS for MongoDB pods launched via the Operator through statefulSet. This makes those pods prone to preemption.
This PR should allow QoS to be set for pods as guaranteed and also allow user to limit resources to the PMM container.


DrFaust92:
done via https://github.com/percona/percona-helm-charts/pull/406

PR #398, opened 2024-10-29
Skip Everest from automated releases

mayankshah1607:
We'd like a separate, manual release process for Everest Helm chart, hence removing it from automated releases


mayankshah1607:
I had already added it to `path-ignore`, but it still tries to release the chart and fails - https://github.com/percona/percona-helm-charts/actions/runs/11568362248/job/32200241354

tplavcic (approved):

PR #400, opened 2024-10-29
Fix Everest chart release workflow

mayankshah1607:
Looks like you need to update Chart.lock after updating the verrion in Chart.yaml

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -36,7 +36,7 @@ jobs:
           fi
 
       - name: Run chart-testing (lint)
-        run: ct lint --config .github/ct.yaml --lint-conf .github/lintconf.yaml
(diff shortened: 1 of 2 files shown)


recharte (approved):

.github/workflows/test.yaml
```diff
@@ -35,7 +36,7 @@ jobs:
           fi
 
       - name: Run chart-testing (lint)
-        run: ct lint --config .github/ct.yaml --lint-conf .github/lintconf.yaml
+        run: ct lint --config .github/ct.yaml --lint-conf .github/lintconf.yaml --excluded-charts=everest
```
mayankshah1607, line 39: The linter won't pass unless I bump the chart version. Everest has a different release process, and doesn't need the version bump, so I'm excluding it for now. I will add a separate CI workflow for Everest only later, and it will cover both testing and linting..

recharte (approved):

PR #401, opened 2024-10-29
EVEREST-1595 | Openshift support

mayankshah1607:


charts/everest/README.md
```diff
@@ -63,11 +63,13 @@ The following table shows the configurable parameters of the Percona Everest cha
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| compatibility.openshift | bool | `false` | If set, enable OpenShift compatibility. |
```
michal-kralik, line 66: It seems the `compatibility.openshift` identifies the namespace where catalog is installed.
What do you think about providing the option to specify the namespace for the catalog instead?

With the current approach it's not very clear what "enable OpenShift compatibility" stands for.
By allowing to specify the catalog namespace we would also provide support for non-openshift clusters which already run OLM.

mayankshah1607: > specify the namespace for the catalog instead?
We already have a flag for that called `olm.namespace`

> provide support for non-openshift clusters which already run OLM.
We can do that by specifying `olm.namespace=<namespace>` and `olm.install=false`

mayankshah1607: > With the current approach it's not very clear what "enable OpenShift compatibility" stands for.

I'm just trying to hide the actual details of what settings need to change to make it work on OpenShift. I know right now there's not much to hide, just need to toggle a namespace, but in the future if we discover that more changes need to be made, then they can simply be toggled on by setting `compatibility.openshift`

> provide support for non-openshift clusters which already run OLM.

Moreover, this scenario is covered as mentioned in the previous comment

michal-kralik (dismissed):

michal-kralik (approved):

PR #402, opened 2024-10-29
Fix Everest chart release workflow

mayankshah1607:
- Use chart releaser `v1.6.0`
- Add step for releasing sub-chart

### DIFF
--- a/.github/workflows/everest-release.yaml
+++ b/.github/workflows/everest-release.yaml
@@ -47,7 +47,15 @@ jobs:
         with:
           version: v3.4.0
 
-      - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.1.0
+      - name: Release chart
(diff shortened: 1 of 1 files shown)


.github/workflows/everest-release.yaml
```diff
@@ -47,7 +47,15 @@ jobs:
...
 
-      - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.1.0
+      - name: Release chart
+        uses: helm/chart-releaser-action@v1.6.0
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Release sub-chart
+        uses: helm/chart-releaser-action@v1.6.0
```
recharte, line 56: Don't we need the CR_TOKEN here as well?

recharte (approved):

PR #416, opened 2024-11-16
EVEREST-1516 | Use VictoriaMetrics Helm chart to install Everest monitoring

mayankshah1607:
We get rid of the OLM way of installing VictoriaMetrics since it requires the use of chart hooks, and also has problems integrating with GitOps tools. This PR uses the victoria metrics Helm chart as a dependency (sub-chart) to install the VM operator.

I've run some tests on this installation of Helm, the VMagent and the metrics continue to work as normal


michal-kralik (approved):

charts/everest/Makefile
```diff
@@ -8,7 +8,6 @@ prepare-chart: deps
 	yq eval -i '.olm.catalogSourceImage = "$(IMAGE_PREFIX)/everest-catalog"' values.yaml
 	yq eval -i '.operator.image = "$(IMAGE_PREFIX)/everest-operator"' values.yaml
 	yq eval -i '(.dependencies[] | select(.name == "everest-db-namespace")).version = "${VERSION}"' Chart.yaml
-	helm dependency update .
```
recharte, line 11: Why did this get removed?

charts/everest/values.yaml
```diff
@@ -101,3 +96,20 @@ kube-state-metrics:
...
+createMonitoringResources: true
+
+# Do not change the following values unless you know what you are doing.
+# @ignore
+monitoring:
+  enabled: true
+  namespaceOverride: everest-monitoring
+  fullnameOverride: vm-operator
+  crds:
+    enabled: false
```
recharte, line 110: Don't we need this?

charts/everest/templates/monitoring-resources.yaml
```diff
@@ -0,0 +1,110 @@
...
+  - replacement: kubelet
+    targetLabel: job
+  scheme: https
+  scrapeTimeout: 5s
+  tlsConfig:
+    caFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+    insecureSkipVerify: true
+---
+apiVersion: operator.victoriametrics.com/v1beta1
+kind: VMPodScrape
```
recharte, line 96: Were you able to confirm that we need this after all?

charts/everest/Makefile
mayankshah1607: This command already runs in `make deps`, so it is redundant here

charts/everest/templates/monitoring-resources.yaml
mayankshah1607: Not yet, but once I have confirmation, I will remove it in a follow-up PR.

For now I'm just moving everything from the everest repo, so that there's track of what we moved/removed and why we did so

recharte (approved):

PR #425, opened 2024-11-22
EVEREST-1511 | support for deploying default DB namespace during installation

mayankshah1607:
This PR brings 2 big improvements:

#### 1. Deploy a default DB namespace along with the core chart installation:
- By default, the `everest` DB namespace will be deployed along with the core chart installation
- The default DB namespace may be skipped from the core chart by specifying `dbNamespace.enabled=false`
- The default DB namespace name may be overriden by specifying `dbNamespace.namespaceOverride`
- By default, all operators are installed. Specific operators may be skipped by specifying `pxc|pg|psmdb=false`

#### 2. No longer need to manually create the namespace object before installing DB namespace

- Earlier you were required to create the namespace beforehand, so that we could attach the `app.kubernetes.io/managed-by=everest` label
- With this PR, the labelling happens automatically as a part of the post-install hook. So you no longer need to create a namespace manually.

See the updated README for full info.


charts/everest/README.md
```diff
@@ -12,51 +12,82 @@ but if you do so, please be aware of the risks.
...
 ```sh
 helm repo add percona https://percona.github.io/percona-helm-charts/
-helm install everest-core percona/everest --namespace everest-system --create-namespace
+helm repo update
 ```
 
-> Note: we currently do not support deploying Everest in a namespace other than `everest-system`.
+### 2. Install Everest
+```sh
+helm repo add percona https://percona.github.io/percona-helm-charts/
```
recharte, line 24:
```suggestion
```

charts/everest/README.md
```diff
@@ -12,51 +12,82 @@ but if you do so, please be aware of the risks.
...
-kind: Namespace
-metadata:
-  name: everest
-  labels:
-    app.kubernetes.io/managed-by: everest
-EOF
-helm install everest percona/everest-db-namespace --namespace everest
+helm install everest \
+    percona/everest-db-namespace \
+    --namespace everest
```
recharte, line 64: Why doesn't this need the `--create-namespace`?

charts/everest/README.md
```diff
@@ -12,51 +12,82 @@ but if you do so, please be aware of the risks.
...
+helm uninstall everest -n <your_db_namespace>
+kubectl delete ns <your_db_namespace>
 ```
 
-Then you can uninstall Everest itself:
+Then run the below command to uninstall Everest from your cluster.
 ```sh
 helm uninstall everest-core -n everest-system
 kubectl delete ns everest-system
 ```
```
recharte, line 84: Maybe we could separate the 2 actions in 2 subsections (5.1 and 5.2) to make it more visible that there are 2 distinct steps depending on the installation.

.github/workflows/everest-pr-checks.yaml
```diff
@@ -83,47 +83,44 @@ jobs:
...
           kubectl wait --for=jsonpath='.status.status'=installed dbengine/percona-postgresql-operator --timeout=600s -n everest
 
-          # Uninstall
-          helm uninstall everest -n everest
-          kubectl delete ns everest
-          kubectl wait --for=delete namespace/everest --timeout=120s
-          helm uninstall everest-core -n everest-system
+          # Provision new DB namespace
+          kubectl create ns everest-db
+          kubectl label ns everest-db app.kubernetes.io/managed-by=everest
```
recharte, line 107: Shouldn't these 2 lines be removed now that we can do this directly from the helm command?

charts/everest/README.md
recharte: I just saw that you have the `charts/everest/charts/everest-db-namespace/templates/namespace.yaml`. Is that why we don't need the `--create-namespace`?

Since you have the hook that labels the namespace couldn't we leverage the `--create-namespace` feature of helm?

mayankshah1607: It needs the flag, I missed it in the docs. Thanks for catching it :)

 > I just saw that you have the charts/everest/charts/everest-db-namespace/templates/namespace.yaml. Is that why we don't need the --create-namespace?

No, that template is rendered only when you deploy the sub-chart with the main chart, since Helm will not create that namespace for you otherwise

charts/everest/README.md
mayankshah1607: Yes, you're right. Fixed it :)

.github/workflows/everest-pr-checks.yaml
mayankshah1607: Yes, you're right. Fixed it :)

recharte (approved):

PR #436, opened 2024-12-03
EVEREST-107 Remove kube-rbac-proxy and update kubebuilder project to go/v4

recharte:
In https://github.com/percona/everest-operator/pull/597 we removed kube-rbac-proxy and migrated to go/v4 project. This commit reflects the manifest changes introduced in that PR.

### DIFF
--- a/charts/everest/templates/everest-operator/backupstorage-editor.clusterrole.yaml
+++ b/charts/everest/templates/everest-operator/backupstorage-editor.clusterrole.yaml
@@ -1,0 +1,23 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: everest-operator-backupstorage-editor-role
+rules:
+- apiGroups:
+  - everest.percona.com
+  resources:
+  - backupstorages
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - everest.percona.com
+  resources:
+  - backupstorages/status
+  verbs:
+  - get
--- a/charts/everest/templates/everest-operator/deployment.yaml
+++ b/charts/everest/templates/everest-operator/deployment.yaml
@@ -19,54 +19,11 @@ spec:
       labels:
         app: everest-operator
     spec:
-      affinity:
-        nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-            - matchExpressions:
-              - key: kubernetes.io/arch
-                operator: In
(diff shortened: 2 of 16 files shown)


charts/everest/templates/everest-operator/backupstorage-editor.clusterrole.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - everest.percona.com
+  resources:
+  - backupstorages/status
+  verbs:
+  - get
```
mayankshah1607, line 23: Don't we need a ClusterRoleBinding to bind all these ClusterRoles to the ServiceAccount?

mayankshah1607 (approved):
LGTM, just one question

charts/everest/templates/everest-operator/deployment.yaml
```diff
@@ -19,54 +19,11 @@ spec:
...
-          allowPrivilegeEscalation: false
-          capabilities:
-            drop:
-            - ALL
-        terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
-      - args:
-        - --health-probe-bind-address={{ .Values.operator.healthProbeAddr }}
-        - --metrics-bind-address={{ .Values.operator.metricsAddr }}
-        {{- if .Values.operator.enableLeaderElection }}
```
mayankshah1607, line 68: This `{{- if ... }}` needs to be added back, that's what the linter complains about..

charts/everest/templates/everest-operator/deployment.yaml
```diff
@@ -19,54 +19,11 @@ spec:
...
+        - --metrics-bind-address=:8443
         - --leader-elect
+        - --health-probe-bind-address=:8081
```
mayankshah1607, line 26: These flags should be placed outside the `if-else` block, since they don't depend on leader election

charts/everest/templates/everest-operator/backupstorage-editor.clusterrole.yaml
recharte: I was actually thinking if should have added these ClusterRoles at all because no binding was generated so I think that they are created thinking about some future service account but aren't directly used by the operator. Do you think we should remove them since they are not being used?

mayankshah1607: Okay.. If they serve no purpose, I guess lets just remove them 😛

I also don't see any harm in keeping them..

mayankshah1607 (approved):

PR #440, opened 2024-12-09
EVEREST-1732 | Fix pre-delete hook (DB resource cleanup)

mayankshah1607:
- update the Job RBAC so that `--wait` works correctly. Currently, without the `get` and `watch` permissions, `--wait` doesn't actually wait for the deletion and fails silently
- add appropriate hook weight so that the DB clean-up always runs first

### DIFF
--- a/charts/everest/charts/common/templates/_db_resources_cleanup.yaml.tpl
+++ b/charts/everest/charts/common/templates/_db_resources_cleanup.yaml.tpl
@@ -11,6 +11,7 @@ metadata:
   annotations:
     "helm.sh/hook": pre-delete
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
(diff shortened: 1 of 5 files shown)


charts/everest/charts/common/templates/_db_resources_cleanup.yaml.tpl
```diff
@@ -56,6 +61,7 @@ metadata:
   annotations:
     "helm.sh/hook": pre-delete
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook-weight": "-1"
```
mayankshah1607, line 64: Helm sorts the hooks based on the ascending order of the weight, so in this case, `-1` runs before all

oksana-grishchenko (approved):

PR #444, opened 2024-12-12
EVEREST-1746 | Fix broken script in pre-upgrade hook

mayankshah1607:
* Does not correctly resolve the CLI binary URL
* Requires new permissions to work correctly

### DIFF
--- a/charts/everest/charts/common/templates/_upgrade_checks.yaml.tpl
+++ b/charts/everest/charts/common/templates/_upgrade_checks.yaml.tpl
@@ -5,6 +5,48 @@
 #
 {{- define "everest.preUpgradeChecks" }}
 {{- $hookName := printf "everest-helm-pre-upgrade-hook" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ $hookName }}
+  namespace: {{ .namespace }}
+  annotations:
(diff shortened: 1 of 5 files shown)


charts/everest/charts/common/templates/_upgrade_checks.yaml.tpl
```diff
@@ -21,18 +63,31 @@ spec:
...
+              fi
+
               VERSION={{ .version }}
               apk add --no-cache --quiet curl
               curl -sSL -o everestctl https://github.com/percona/everest/releases/download/v${VERSION}/everestctl-${OS}-${ARCH}
               chmod -R 777 ./everestctl
               
               echo "Checking requirements for upgrade to version ${VERSION}"
-              ./everestctl upgrade --dry-run --version-metadata-url={{ .versionMetadataURL }}
+              ./everestctl upgrade --in-cluster --dry-run --version-metadata-url={{ .versionMetadataURL }} --kubeconfig=""
```
mayankshah1607, line 87: There is an issue in the CLI where it tries to read kubeconfig from `~/.kube/config` even though `--in-cluster` is specified. Setting `--kubeconfig=""` resolves it. 

I know its a bit weird, but I'm doing it this way for now to unblock testing.. I will fix it in the CLI later

PR #452, opened 2024-12-18
EVEREST-1762 | Prevent Helm from deleting Subscriptions

mayankshah1607:
* Adding the `helm.sh/resource-policy: keep` annotation to a resource will ensure that Helm does not remove it during upgrade/uninstall
* The pre-delete hook takes care of deleting the subscription during uninstall.

This way, a Subscription may not be removed after creating it


oksana-grishchenko (approved):
🚀

recharte (dismissed):

oksana-grishchenko (approved):

PR #467, opened 2025-01-14
EVEREST-107 | add docs for openshift

mayankshah1607:
- also adds an additional manifest for openshift compatibly


charts/everest/templates/ksm-openshift-clusterrolebinding.yaml
```diff
@@ -0,0 +1,17 @@
+# We need a separate ClusterRoleBinding for OpenShift, as the one that comes pre-intalled cannot be bound to the ServiceAccount used
```
tplavcic, line 1:
```suggestion
# We need a separate ClusterRoleBinding for OpenShift, as the one that comes pre-installed cannot be bound to the ServiceAccount used
```

recharte (approved):

tplavcic (approved):

PR #475, opened 2025-01-23
EVEREST-1837 fix operator.env and server.env value types

recharte:


recharte:
https://github.com/percona/everest/issues/919

Issue #480, opened 2025-01-31
Unable to deploy standby cluster using helm-chart due to missing PG Custom Resource options


elsayedmt:
Background:

I am trying to deploy a PG cluster using the guide https://docs.percona.com/percona-operator-for-postgresql/2.0/update-standby.html to migrate a cluster managed by version 1.x of the operator to version 2.x using the standby approach. My configuration works fine but when I set standby: true, the operator complains that repoName isn't specified by is required. 

Issue:

The standby.repoName and other fields are not passed into the PerconaPGCluster CR. See here https://docs.percona.com/percona-operator-for-postgresql/2.0/operator.html#standbyenabled

elsayedmt:
Closes #481

PR #481, opened 2025-01-31
pg-db - Added missing PerconaPGCluster .spec.standby. fields

elsayedmt:
This PR should fix the missing fields in .spec.standby by passing them all-through leaving the validation up to the operator


elsayedmt:
I also noticed that these fields are present and commented in the values.yaml file. My guess is that this use case + deployment method isn't very common or that people customize the helm chart locally and use it (such was the case for us)

elsayedmt:
Closes #480

jvpasinatto (approved):

jvpasinatto:
Hi @elsayedmt, thanks for the contribution!

PR #483, opened 2025-02-04
PMM-13725 Allow specifying namespace.

BupycHuk:


charts/pmm/values.yaml
```diff
@@ -211,7 +215,7 @@ extraLabels: {}
 ## @param serviceAccount.name Name of the service account to use. If not set and create is true, a name is generated using the fullname template.
 ##
 serviceAccount:
-  create: true
+  create: false
```
mayankshah1607, line 218: Not sure if this change is intentional, everything else looks good

mayankshah1607 (approved):
LGTM, thanks for the quick turnaround

idoqo (approved):

charts/pmm/values.yaml
BupycHuk: yes, that's intentional

PR #491, opened 2025-02-19
Pg db metadata passthrough

elsayedmt:
Why

I am trying to setup open telemetry collector sidecars on the postgres instance pods and hence need to set Pod metadata.annotations on the postgres instances.

How

Adding the metadata fields supported by the operator as in here https://github.com/percona/percona-postgresql-operator/blob/main/pkg/apis/pgv2.percona.com/v2/perconapgcluster_types.go#L52

What

modification to pg-db chart to passthrough the meta data fields + associated documentation


jvpasinatto (approved):

jvpasinatto:
Hi @elsayedmt 
Thanks for the contribution!

PR #497, opened 2025-03-07
K8SPSMDB-1219: Add main storage example

egegunes:


jvpasinatto (changes requested):
please add crd and change branch to [K8SPSMDB-1265](https://github.com/percona/percona-helm-charts/tree/K8SPSMDB-1265)

PR #500, opened 2025-03-14
EVEREST-107 update VictoriaMetrics operator to v0.54.1

recharte:


charts/everest/values.yaml
```diff
@@ -152,7 +152,7 @@ monitoring:
   namespaceOverride: everest-monitoring
   fullnameOverride: vm-operator
   crds:
-    enabled: false
+    enabled: true
```
oksana-grishchenko, line 155: Could you please elaborate more on why do we need to change this value?

recharte: I'm going to ask @mayankshah1607 to chime in because he can give a better answer than me :see_no_evil:

recharte: Without it we get the following error while upgrading:
```
❌  Upgrading Custom Resource Definitions
❌ could not render Helm templates: execution error at (everest/charts/monitoring/templates/crd.yaml:6:6): When CRDs are disabled (`crds.enabled: false`) you need to disable plain CRD rendering (`crds.plain: false`)
```

mayankshah1607: At least based on what I understand, somewhere between `0.37` and `0.41`, the meaning of this `enabled` flag was changed.. Earlier you could set `enabled=false` and have plain CRDs rendered, but with this one, you need to first have `enabled=true` (which gives us templated CRDs by default), and then `crds.plain=true` (to have standard CRDs)

mayankshah1607 (approved):

charts/everest/values.yaml
oksana-grishchenko: Ok, got it, thank you folks for the elaboration. 

It's just this note
```
# Do not change the following values unless you know what you are doing.
```
made me concerned since I wasn't sure what is the purpose of the change

oksana-grishchenko (approved):

PR #510, opened 2025-03-27
K8SPG-704: Add create replica methods

egegunes:


charts/pg-db/templates/cluster.yaml
```diff
@@ -518,6 +518,10 @@ spec:
       targetInstance: {{ .Values.patroni.switchover.targetInstance }}
       {{- end }}
     {{- end }}
+    {{- if .Values.patroni.createReplicaMethods }}
+    createReplicaMethods:
```
jvpasinatto, line 522: please add this field to readme

jvpasinatto (changes requested):

charts/pg-db/templates/cluster.yaml
egegunes: https://github.com/percona/percona-helm-charts/pull/510/commits/8f7e1c9000b94c4492119adb848bd8cb2fcdec7f

PR #516, opened 2025-04-02
K8SPG-710: Add field to enable/disable backups

egegunes:
Added a new field `backups.enabled`. By default it's true and if this key does not exist in `values.yaml` backups should be enabled too. If you set it to `false`, backup section shouldn't be generated in cr.yaml


charts/pg-db/templates/cluster.yaml
```diff
@@ -365,6 +365,9 @@ spec:
     postgresParams: {{ .Values.pmm.postgresParams}}
 
   backups:
+{{ if (hasKey .Values.backups "enabled") }}
```
valmiranogueira: To avoid empty lines on the template try:
 {{ - if

valmiranogueira: Also, all other .Values.backups (like image ...) remains required, probably the end of this if will be after entire backups section

eleo007 (approved):

PR #523, opened 2025-04-10
fix chart for using watchAllNamespaces and correct Role/ClusterRole management

ksemele:
I suggest some changes similar to pg-operator.
It actualize Role and ClusterRole management and allow using `watchAllNamespaces` as in other Percona operators


dbazhenov:
Hi, thank you!

I want to add a link to the issue that the PR solves
https://github.com/percona/percona-server-mysql-operator/issues/886

Hopefully the reviewers and QA will check your contribution soon.

ksemele:
Should I do something to pass the `lint-test` step?

ksemele:
Hello everyone,
Could I kindly request you to review this PR carefully?
Could you please provide some guidance on areas where I might improve it?
@hors @valmiranogueira, please?

charts/ps-operator/templates/deployment.yaml
```diff
@@ -31,11 +31,19 @@ spec:
...
           value: "{{ .Values.env.logLevel }}"
+        {{- if and (not .Values.watchNamespace) (not .Values.watchAllNamespaces) }}
         - name: WATCH_NAMESPACE
           valueFrom:
             fieldRef:
               apiVersion: v1
               fieldPath: metadata.namespace
+        {{- else if .Values.watchNamespace }}
+        - name: WATCH_NAMESPACE
+          value: {{ .Values.watchNamespace }}
```
valmiranogueira: Please consider `.Release.Namespace` by default, when `watchNamespace` is not defined.

charts/ps-operator/README.md
```diff
@@ -39,6 +39,8 @@ The chart can be customized using the following configurable parameters:
 | `env.logStructured`     | Enable JSON format for logs                                                                    | `false`                                 |
 | `env.logLevel`          | Set appropriate log level (INFO, DEBUG, ERROR)                                                 | `INFO`                                  |
 | `disableTelemetry`      | Disable sending PS Operator telemetry data to Percona                                          | `false`                                 |
+| `watchNamespace`     | Set this variable if the target cluster namespace differs from operators namespace |
```
valmiranogueira: Please align columns here, default value column is missing

valmiranogueira (commented):
Please bump the chart min version (0.11.1 for example)

charts/ps-operator/templates/deployment.yaml
ksemele: There is no reason to do that because it's already its.
Here is:
`fieldPath: metadata.namespace`
what is in reality:
`namespace: {{ .Release.Namespace }}`

It's the original behavior of this chart

ksemele: But if we want to change this behavior and make it more explicit... maybe it should be done, but in a separate MR, because then we need to fix this behavior here as well: https://github.com/percona/percona-helm-charts/blob/main/charts/pg-operator/templates/deployment.yaml#L56
Then we will get almost identical structures in the charts.

valmiranogueira: Yeap, but if you look into https://github.com/percona/percona-helm-charts/blob/main/charts/pxc-operator/templates/deployment.yaml#L58 that looks a bit more simple (one less condition), but in fact the default is `metadata.namespace` as you mentioned.

ksemele: Ah, I see. Okay, this looks better!

PR #526, opened 2025-04-11
K8SPG-726 update charts

nmarukovich:


charts/pg-db/templates/cluster.yaml
```diff
@@ -208,6 +208,10 @@ spec:
       {{- range $sidecar := $instance.sidecars }}
       - name: {{ $sidecar.name }}
         image: {{ $sidecar.image }}
+        command: {{ $sidecar.command }}
```
valmiranogueira: After some testing, I found that this would cause a crashloop cause it needs to be converted to an array

charts/pg-db/templates/cluster.yaml
```diff
@@ -208,6 +208,10 @@ spec:
...
+        {{- if $sidecar.securityContext }}
+        securityContext: {{ $sidecar.SecurityContext | toYaml | indent 8 }}
```
valmiranogueira: SecurityContext is securityContext in values file, this probably must match

jvpasinatto (approved):

PR #527, opened 2025-04-12
pg-db: allow customization of pmm secret

fergusean:
The pg-db values file includes a key for the PMM secret, but the chart doesn't reference it.  We have a use case for setting the secret name, rather than having it auto-derived from the template.  Defaults to original calculation method & comments out default in values file to avoid breaking existing installations.


gkech:
seems that we already introduced the aforementioned changes, check this pr: https://github.com/percona/percona-helm-charts/pull/551

gkech:
we are closing this, thanks a lot

PR #531, opened 2025-04-17
EVEREST-1520 Implement PodSchedulingPolicy feature

maxkondr:
Everest PR - https://github.com/percona/everest/pull/1322
Everest Operator PR - https://github.com/percona/everest-operator/pull/737


charts/everest/charts/common/templates/_psp_cleanup.yaml.tpl
```diff
@@ -0,0 +1,77 @@
...
+  template:
+    spec:
+      containers:
+        - image: bitnami/kubectl:latest
+          name: {{ $hookName }}
+          command:
+            - /bin/sh
+            - -c
+            - |
+              for pspName in everest-default-pxc everest-default-psmdb everest-default-postgresql
```
oksana-grishchenko, line 67: If at some point we're going to provide more defaults or the default name gets changed we would need to change this loop too. Would it be too bad if we patch all the psps? Besides, users might have applied the `everest.percona.com/readonly-protection` finalizer to other psp manually to protect them.

maxkondr: done

mayankshah1607 (approved):

oksana-grishchenko (approved):

charts/everest/charts/common/templates/_assign_default_psp.yaml.tpl
```diff
@@ -0,0 +1,150 @@
...
+#
+{{- define "everest.assignDefaultPsp" }}
+{{- $hookName := printf "everest-helm-psp-post-upgrade-hook" }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ $hookName }}
+  namespace: {{ .namespace }}
+  annotations:
+    "helm.sh/hook": post-upgrade
```
mayankshah1607, line 12: Did you test that this actually works (does not result in restarts?)

This hook is run _after_ the upgrade is complete, meaning that everest-operator would have upgraded to 1.7.0 before this hooks run. Wouldn't it have had the time to run at least one reconciliation where it changes the affinity to `nil`, before this hook kicks in?

charts/everest/charts/common/templates/_assign_default_psp.yaml.tpl
```diff
@@ -0,0 +1,150 @@
...
+          command:
+            - /bin/sh
+            - -c
+            - |
+              for nsName in `kubectl get namespace -lapp.kubernetes.io/managed-by=everest --no-headers -o jsonpath='{.items[*].metadata.name}'`
+              do
+                  for dbName in `kubectl get databaseclusters -n $nsName -o jsonpath='{.items[?(@.spec.engine.type=="postgresql")].metadata.name}'`
+                  do
+                    kubectl patch -n $nsName  databasecluster/$dbName -p '{"spec":{"podSchedulingPolicyName": "everest-default-postgresql"}}' --type=merge
+                  done
```
mayankshah1607, line 78: Looks like we're unconditionally patching all DBs? What happens when we upgrade to 1.8.0? This will reset all DBs' referenced psps to the default?

charts/everest/charts/common/templates/_assign_default_psp.yaml.tpl
```diff
@@ -0,0 +1,150 @@
...
+            - /bin/sh
+            - -c
+            - |
+              for nsName in `kubectl get namespace -lapp.kubernetes.io/managed-by=everest --no-headers -o jsonpath='{.items[*].metadata.name}'`
+              do
+                  for dbName in `kubectl get databaseclusters -n $nsName -o jsonpath='{.items[?(@.spec.engine.type=="psmdb")].metadata.name}'`
+                  do
+                    kubectl patch -n $nsName  databasecluster/$dbName -p '{"spec":{"podSchedulingPolicyName": "everest-default-psmdb"}}' --type=merge
+                  done
+              done
```
mayankshah1607, line 110: Do we actually need 3 jobs? Can't we combine the script to run 1 job?

charts/everest/charts/common/templates/_assign_default_psp.yaml.tpl
maxkondr: This hook needs only once during update to 1.7.0. After that we will remove it

charts/everest/charts/common/templates/_assign_default_psp.yaml.tpl
maxkondr: we can, just wanted to have a separate tasks, will rework

charts/everest/templates/hooks.yaml
```diff
@@ -1,3 +1,8 @@
 {{- if .Values.upgrade.preflightChecks }}
 {{- include "everest.preUpgradeChecks" (dict "namespace" (include "everest.namespace" .) "version" .Chart.Version "versionMetadataURL" (include "everest.versionMetadataURL" .)) }}
 {{- end }}
+---
+{{- include "everest.pspCleanup" (dict "namespace" (include "everest.namespace" .)) }}
+---
+{{- include "everest.assignDefaultPsp" (dict "namespace" (include "everest.namespace" .)) }}
```
mayankshah1607, line 7: Since we should run this only in 1.7.0 ... I know we plan to remove it in 1.8.0, but adding this condition won't harm, in case we forget to. It also acts as self-explanatory documentation 😅 
```suggestion
{{- if eq .Chart.Version "1.7.0" }}
{{- include "everest.assignDefaultPsp" (dict "namespace" (include "everest.namespace" .)) }}
{{- end }}
```

charts/everest/charts/common/templates/_assign_default_psp.yaml.tpl
mayankshah1607: https://github.com/percona/percona-helm-charts/pull/531/files#r2081140998

charts/everest/templates/hooks.yaml
maxkondr: Done

charts/everest/charts/common/templates/_assign_default_psp.yaml.tpl
maxkondr: updated, now default PSPs are assigned without restarting DB clusters

charts/everest/charts/common/templates/_assign_default_psp.yaml.tpl
maxkondr: fixed

charts/everest/templates/hooks.yaml
```diff
@@ -1,3 +1,27 @@
...
+{{- if  .Release.IsInstall }}
+{{- include "everest.createDefaultPsp" . }}
+{{- end }}
+
+{{- /*
+# During the upgrade to Everest v1.7.0 we need to:
+#  - create default pod scheduling policies.
+#  - assign default pod scheduling policies to all existing DB clusters.
+ */}}
+{{- /* if and (.Release.IsUpgrade) (eq .Chart.Version "1.7.0") */}}
```
mayankshah1607, line 19: Should this be uncommented?

charts/everest/templates/hooks.yaml
```diff
@@ -1,3 +1,27 @@
 {{- if .Values.upgrade.preflightChecks }}
 {{- include "everest.preUpgradeChecks" (dict "namespace" (include "everest.namespace" .) "version" .Chart.Version "versionMetadataURL" (include "everest.versionMetadataURL" .)) }}
 {{- end }}
+---
+{{- /*
+# --- Pod Scheduling Policies
+# During the installation of Everest v1.7.0 we need to create default pod scheduling policies.
+*/}}
+
+{{- if  .Release.IsInstall }}
```
mayankshah1607, line 10: Why only during installation?

charts/everest/charts/common/templates/_create_default_psp.yaml.tpl
```diff
@@ -0,0 +1,91 @@
...
+{{- define "everest.createDefaultPsp" }}
+apiVersion: everest.percona.com/v1alpha1
+kind: PodSchedulingPolicy
+metadata:
+  name: everest-default-pxc
+  finalizers:
+    - everest.percona.com/readonly-protection
+  annotations:
+      "helm.sh/hook": post-install,pre-upgrade
+      "helm.sh/resource-policy": keep
```
mayankshah1607, line 12: Why do we need this `"helm.sh/resource-policy": keep` annotation?

charts/everest/templates/hooks.yaml
maxkondr: yes, just before we release v1.7.0. Otherwise it is not executed during upgrading v1.6.0 to Feature Build

charts/everest/charts/common/templates/_create_default_psp.yaml.tpl
mayankshah1607: What I meant to ask was - why should these PSP never be deleted by Helm in the future? 🤔

charts/everest/templates/hooks.yaml
mayankshah1607: I saw that, what I meant to ask that if it needs to be created during install and upgrade (basically always), why even have this condition?

charts/everest/charts/common/templates/_create_default_psp.yaml.tpl
maxkondr: ah, I ment this is protection from being removed by hook controller. 
We remove these PSPs in the `delete-hook` manually

charts/everest/templates/hooks.yaml
maxkondr: because in the next release the whole block 
```
{{- /* if and (.Release.IsUpgrade) (eq .Chart.Version "1.7.0") */}}
...
{{ - end }}
```

will be deleted as soon as it won't be needed any more

mayankshah1607 (dismissed):

mayankshah1607 (approved):

PR #535, opened 2025-04-23
K8SPXC-1615 custom name for pmm

nmarukovich:


jvpasinatto (changes requested):
Please change base to 1.19.0 release branch and add this field in cluster.yaml

PR #536, opened 2025-04-24
EVEREST-2014 | Set DISABLE_TELEMETRY env var in server deployment

mayankshah1607:


charts/everest/README.md
```diff
@@ -191,6 +191,6 @@ The following table shows the configurable parameters of the Percona Everest cha
 | server.service.name | string | `"everest"` | Name of the service for everest |
 | server.service.port | int | `8080` | Port to expose on the service. |
 | server.service.type | string | `"ClusterIP"` | Type of service to create. |
-| telemetry | bool | `true` | If set, enabled sending telemetry information. |
+| telemetry | bool | `false` | If set, enabled sending telemetry information. |
```
oksana-grishchenko, line 194: Isn't it a bit confusing that here in the readme we define that the telemetry is false by default, however in the release builds it is actually enabled 🤔
It can be treated as the default readme for the latest dev build though then we are good.

oksana-grishchenko (dismissed):

charts/everest/README.md
mayankshah1607: Yeah, I agree it can be confusing.. Maybe we can add a comment there saying it will be different in the prod builds

maxkondr (approved):

oksana-grishchenko (approved):

PR #537, opened 2025-04-29
EVEREST-2018 | Fix port-forward instructions for TLS

mayankshah1607:


oksana-grishchenko (approved):
Note: the `everest` repo also has the mention of port-forwarding to 8080 in the CLI post-install message and in README.md, should we update the information there as well?

mayankshah1607:
@oksana-grishchenko yes, already created a PR - https://github.com/percona/everest/pull/1340

PR #544, opened 2025-05-15
EVEREST-1999, EVEREST-2002 | Helm chart improvements

mayankshah1607:
The following changes are made to the Everest Helm chart:

#### 1. Introduce everest-crds sub-chart for managing CRDs separately
* Everest CRDs have been moved into a dedicated everest-crds sub-chart.
* The CRD files in this sub-chart are symlinked into the main everest/crds/ directory, so Helm still installs them automatically during a fresh install.
* Helm handles these symlinks at release time.
* **Users must manually upgrade the CRDs using the everest-crds sub-chart before upgrading the main everest chart.**

#### 2. Add guardrails to ensure CRDs are upgraded first
* All Everest CRDs now include a label `everest.percona.com/version`, indicating the chart version they were installed from.
* A static check has been added in the Everest chart templates to ensure all CRDs have the expected label version before allowing the upgrade.
* Note that this is check is done entirely with Helm functions, so no hooks are added.

#### 3. OLM Manifests Moved to a Separate Chart
* All OLM-related manifests have been moved to a dedicated Helm chart maintained under the Percona fork of OLM: https://github.com/percona/operator-lifecycle-manager/tree/master
* The Everest Helm chart now includes this OLM chart as a dependency.
* Note: When installing on OpenShift, you must explicitly set olm.install=false.
A static validation (not hook-based) has been added to ensure that compatibility.openshift=true and olm.install=true are never set at the same time, as this combination is unsupported.

### Important Note

* Please see the newly updated upgrade documentation since there may be some new flags you would have to add for installing the CRD chart for the first time


charts/everest/Makefile
```diff
@@ -42,18 +54,29 @@ deps: add-repos
...
 EVEREST_REPO_URL ?= https://github.com/percona/everest-operator
 CRD_DIR ?= config/crd
 CRD_VERSION ?= main
 .PHONY: crds-gen
 crds-gen:
-	docker run --rm registry.k8s.io/kustomize/kustomize:v5.0.0 build $(EVEREST_REPO_URL)/$(CRD_DIR)?ref=$(CRD_VERSION) > crds/everest.yaml
+	EVEREST_REPO_URL=$(EVEREST_REPO_URL) \
+	CRD_DIR=$(CRD_DIR) \
+	CRD_VERSION=$(CRD_VERSION) \
+	VERSION=$(VERSION) \
```
oksana-grishchenko, line 73: are those explicitly set here for visibility purposes only (so that we saw all the vars that matter for the command) or there is also some other reason?

charts/everest/charts/everest-crds/hack/crd.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+
+import sys
+import argparse
+import yaml
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--output", required=True, help="Output directory for split CRD files")
+    parser.add_argument("--labels", help="Comma-separated labels to add (e.g. key1=val1,key2=val2)")
```
oksana-grishchenko: does it add the new labels or rewrite the whole labels section? 
If it only adds the new labels, then why in the Makefile we pass the whole list of labels including `everest.percona.com/crd=true` which is already defined in the templates. 
If it rewrites the section, then we need to update help and mention it explicitly to avoid misusage

charts/everest/README.md
```diff
@@ -211,6 +226,7 @@ The following table shows the configurable parameters of the Percona Everest cha
 | server.tls.secret.certs | object | `{"tls.crt":"","tls.key":""}` | Use the specified tls.crt and tls.key in the Secret. If unspecified, the server creates a self-signed certificate (not recommended for production). |
 | server.tls.secret.name | string | `"everest-server-tls"` | Name of the Secret containing the TLS certificates. This Secret is created if tls.enabled=true and certificate.create=false. |
 | telemetry | bool | `false` | If set, enabled sending telemetry information. In production release, this value is `true` by default. |
+| upgrade.crdChecks | bool | `true` | If set, ensures that CRDs are upgraded first. |
```
oksana-grishchenko: > If set, ensures that CRDs are upgraded first.

A bit misleading. The default value is `true` which means that even if the value is not set, the check is going to happen anyway. Is it?

charts/everest/Makefile
```diff
@@ -32,6 +43,7 @@ add-repos:
 	$(HELM) repo add prometheus-community https://prometheus-community.github.io/helm-charts
 	$(HELM) repo add vm https://victoriametrics.github.io/helm-charts
 	$(HELM) repo add percona https://percona.github.io/percona-helm-charts/
+	$(HELM) repo add percona-olm https://percona.github.io/operator-lifecycle-manager/
```
maxkondr: why is a separate repo required?

charts/everest/Makefile
```diff
@@ -42,18 +54,29 @@ deps: add-repos
 docs-gen:
 	docker run --rm -v "$(PWD)/:/helm-docs" -u $(shell id -u) jnorwood/helm-docs:v1.14.2
 
+link-crds:
```
maxkondr, line 58:
```suggestion
.PHONY: link-crds
link-crds:
```

charts/everest/charts/everest-crds/hack/crd.py
oksana-grishchenko, line 1: the folder name "hack" bugs me a bit, it seems like something temporary, a quick and dirty fix. While in fact it contains a script to update crds templates which is going to be integrated into the release process. Should it be just "scripts"?

charts/everest/Makefile
```diff
@@ -42,18 +54,29 @@ deps: add-repos
...
+		echo "Linked $$crd → crds/"; \
+	done
 
 EVEREST_REPO_URL ?= https://github.com/percona/everest-operator
 CRD_DIR ?= config/crd
 CRD_VERSION ?= main
 .PHONY: crds-gen
 crds-gen:
-	docker run --rm registry.k8s.io/kustomize/kustomize:v5.0.0 build $(EVEREST_REPO_URL)/$(CRD_DIR)?ref=$(CRD_VERSION) > crds/everest.yaml
+	EVEREST_REPO_URL=$(EVEREST_REPO_URL) \
```
maxkondr, line 70: do we really need EVEREST_REPO_URL and CRD_DIR ENVs? have we ever changed them?

charts/everest/charts/everest-crds/Makefile
```diff
@@ -0,0 +1,26 @@
...
+		exit 1; \
+	fi
+	@for crd in templates/*.yaml; do \
+		yq eval '.metadata.labels["everest.percona.com/version"] = "$(VERSION)"' $$crd -i; \
+	done
+
+EVEREST_REPO_URL ?= https://github.com/percona/everest-operator
+CRD_DIR ?= config/crd
+CRD_VERSION ?= main
+VERSION ?= 0.0.0
```
oksana-grishchenko: do we need to define `VERSION ?= 0.0.0` twice?

charts/everest/charts/everest-crds/Makefile
```diff
@@ -0,0 +1,26 @@
...
+release:
```
maxkondr, line 3: .PHONY

charts/everest/charts/everest-crds/Makefile
```diff
@@ -0,0 +1,26 @@
...
+	@if [ -z "$(VERSION)" ]; then \
+		echo "Error: VERSION is not set."; \
+		exit 1; \
+	fi
+	@for crd in templates/*.yaml; do \
+		yq eval '.metadata.labels["everest.percona.com/version"] = "$(VERSION)"' $$crd -i; \
+	done
+
+EVEREST_REPO_URL ?= https://github.com/percona/everest-operator
+CRD_DIR ?= config/crd
```
maxkondr, line 13: do we need these ENVs?

charts/everest/charts/everest-crds/Makefile
```diff
@@ -0,0 +1,26 @@
...
+.PHONY: crds-gen
+crds-gen:
+	@crd_tmp_file="crd.tmp.yaml"; \
+	docker run --rm registry.k8s.io/kustomize/kustomize:v5.0.0 \
+		build $(EVEREST_REPO_URL)/$(CRD_DIR)?ref=$(CRD_VERSION) > $(CRD_SUBCHART_DIR)/$$crd_tmp_file; \
+	docker run --rm -i \
+		-v "$(CRD_SUBCHART_DIR)":/workspace \
+		-w /workspace \
+		python:3.11-slim \
+		/bin/sh -c "pip install pyyaml && cat $$crd_tmp_file | python3 ./hack/crd.py --output=templates/ --labels='everest.percona.com/crd=true,everest.percona.com/version=$(VERSION)'"; \
```
maxkondr: can't we achieve the same with kustomize?

charts/everest/templates/check-crd.yaml
```diff
@@ -0,0 +1,22 @@
+# This template checks if all Everest CRDs are up-to-date during a Helm upgrade.
+# - It looks for CRDs with the label "everest.percona.com/crd".
+# - If no such CRDs are found, it fails the upgrade with an error message. Typically this may only ever haappen once, that is, when the chart is upgraded to the
```
maxkondr:
```suggestion
# - If no such CRDs are found, it fails the upgrade with an error message. Typically this may only ever happen once, that is, when the chart is upgraded to the
```

charts/everest/templates/check-crd.yaml
```diff
@@ -0,0 +1,22 @@
...
+#   first version that includes these new labels on the CRDS.
+# - If the CRDs are found but their version (label "everest.percona.com/version") does not match the current chart version, it also fails the upgrade.
+# This ensures that CRDs are upgraded before the Helm release is upgraded.
+
+{{- if (and .Release.IsUpgrade .Values.upgrade.crdChecks )}}
+{{- $found := false }}
+{{- range $index, $crd := (lookup "apiextensions.k8s.io/v1" "CustomResourceDefinition" "" "" ).items }}
```
maxkondr, line 10: where is this lookup performed? in K8S directly?

charts/everest/values.yaml
```diff
@@ -254,3 +259,9 @@ pmm:
   # -- If set, deploys PMM in the release namespace.
   enabled: false
   nameOverride: pmm
+# DO NOT CHANGE.
+# We set this to `false` to prevent the CRDs from being rendered as a template.
+# The CRDs are already installed as a part of the crds/ directory via the sub-chart.
+# @ignore
+everest-crds:
```
maxkondr, line 296: if this var shall not be changed - can we avoid setting it in config at all?

charts/everest/templates/check-crd.yaml
```diff
@@ -0,0 +1,22 @@
...
+  {{- if (and (hasKey $crd "metadata") (hasKey $crd.metadata "labels") (hasKey $crd.metadata.labels "everest.percona.com/crd")) }}
+    {{- $found = true }}
+    {{- $version := index $crd.metadata.labels "everest.percona.com/version" }}
+    {{- if ne $version $.Chart.Version }}
```
oksana-grishchenko: Here we assume that if there is the "everest.percona.com/crd" then there should be the "everest.percona.com/version" as well. Should we also check the presenсe of the "everest.percona.com/version" label before trying to get the index? If the label does not exist, could it cause any problems when comparing the result with the Chart.Version in the template?

charts/everest/charts/everest-crds/Makefile
mayankshah1607: This command is splitting the big YAML into each individual file. How will we achieve this with kustomize?

charts/everest/Makefile
mayankshah1607: Same reason as `kube-state-metrics` and `victoriametrics-operator` -- they are dependencies. The only reason for keeping OLM manifests in this chart was that there is no official OLM Helm chart.

Since we now run our own fork of OLM, and have a repo for it, this gives us an opportunity to clean-up our Everest chart and move OLM there so that it is treated as an actual dependency and not a core Everest manifest ... I don't see a benefit in keeping it here when we have an opportunity to keep it neat

charts/everest/templates/check-crd.yaml
mayankshah1607: Yes

charts/everest/values.yaml
mayankshah1607: Unfortunately we can't get rid of it (I would have already if it were possible) ... we are doing a workaround to get CRDs to work with our chart a certain way, and as a consequence we need to set this value (even though never used) so that CRDs are not rendered twice (once from templates and once from the sub-chart)

charts/everest/README.md
mayankshah1607: Why is it misleading? Yes, by default we will always check that crds are upgraded first .... if you explicitly set it to false, then it will not check..

charts/everest/charts/everest-crds/Makefile
mayankshah1607: ah, didn't realise we are already setting it

charts/everest/Makefile
mayankshah1607: We haven't changed them, but I also don't want it hard-coded in the commands. Is it really an issue that they're there? If we don't want to set it, we can just choose to ignore it ..

charts/everest/charts/everest-crds/Makefile
mayankshah1607: We don't need them but I also don't want to hard-code them in the commands, just to keep the Makefile clean

charts/everest/charts/everest-crds/Makefile
mayankshah1607: Oh nevermind, you're right. It has a flag `--output` that will write to separate files directly. 

thanks for the suggestion, updated

charts/everest/charts/everest-crds/hack/crd.py
mayankshah1607: updated

charts/everest/charts/everest-crds/hack/crd.py
mayankshah1607: I updated the script after Max's suggestions.

> including everest.percona.com/crd=true which is already defined in the templates.

It is not, the files are first generated from kustomize without any labels, and then these are added later so that we can perform the CRD check

charts/everest/Makefile
mayankshah1607: We need to set them because we are calling another Make command

charts/everest/templates/check-crd.yaml
mayankshah1607: Both labels are always set together, so I don't think there will ever be a case when the 'crd' label is there but not the 'version'.. Nevertheless, I added a check anyway since it is pretty straightforward to do

charts/everest/Makefile
maxkondr: aha, I didn't;t know that we have a chart for OLM, now it is clear, thanks

charts/everest/Makefile
maxkondr: each line in Makefile is executed in a separate sub-shell (except the cases when we explicitly tell Make to do in one shell), so that in order to bypass ENVs to a another sub-shell we need to define them

maxkondr (dismissed):

charts/everest/README.md
oksana-grishchenko: I mean, it says "_if set_, ensures blablabla" but even if _not set_  the check _will happen anyway_ bc the default value is true.

oksana-grishchenko (dismissed):

charts/everest/README.md
mayankshah1607: Ah, okay, I see what you mean.

We can change it to `upgrade.skipCRDChecks` - If set, skips checking that CRDs are upgraded first. WDYT? Any other suggestion?

oksana-grishchenko: Yes, I like this idea, `skipCRDChecks` with default `false` would be more straightforward I think

mayankshah1607: As discussed on Slack, updated the comment instead. Please check

maxkondr (approved):

oksana-grishchenko (approved):

PR #545, opened 2025-05-20
PMM-13997 Update PMM chart to handle namespaces only on install

BupycHuk:
Added conditional checks to ensure namespace-related definitions are applied only during chart installation. This prevents unnecessary updates during upgrades and adheres to Helm best practices. Bumped chart version to 1.4.5 for the release.


JiriCtvrtka (dismissed):

charts/pmm/templates/_helpers.tpl
```diff
@@ -37,6 +37,47 @@ Allows overriding the install namespace in combined charts.
...
+Determines whether to include namespace specification.
+Includes namespace for:
+- Fresh installs
+- Upgrades from chart version 1.4.1+ (detected by existing namespace field)
+*/}}
+{{- define "pmm.includeNamespace" -}}
+{{- $shouldInclude := true -}}
+{{- if .Release.IsInstall -}}
+  {{- $shouldInclude = true -}}
+{{- else -}}
```
ademidoff:
```suggestion
{{- if !.Release.IsInstall -}}
```

ademidoff: `$shouldInclude is already true`, so no need for the `else` part.

JiriCtvrtka (approved):

PR #556, opened 2025-05-28
PG-DB - Prevent Template Error When Using Custom TLS and Not 'customRootCATLSSecret'

ryanmrodriguez:
When either `customTLSSecret.name` or `customReplicationTLSSecret.name` are provided, `.Values.customRootCATLSSecret.name` is evaluated in an if condition, but no default is provided for 'customRootCATLSSecret', leading to this error:

```sh
Error: template: ...: executing "...yaml" at <.Values.customRootCATLSSecret.name>: nil pointer evaluating interface {}.name
```

Current defaults:
```yaml
# customRootCATLSSecret:
#   name: cluster1-ca-cert
#   items:
#     - key: "tls.crt"
#       path: "root.crt"
#     - key: "tls.key"
#       path: "root.key"
customTLSSecret:
  name: ""
customReplicationTLSSecret:
  name: ""
```

---

On a side-note, I think all of these options are not technically mutually exclusive? Therefore the entry conditional seems like it could be adjusted: `{{- if or (.Values.customTLSSecret.name) (.Values.customReplicationTLSSecret.name) }}` ?


gkech:
created a ticket: https://perconadev.atlassian.net/browse/K8SPG-1037

PR #562, opened 2025-06-03
Remove PSP hook during upgrade

maxkondr:
https://perconadev.atlassian.net/browse/EVEREST-2093
https://perconadev.atlassian.net/browse/EVEREST-2086


charts/everest/Chart.lock
```diff
@@ -1,10 +1,10 @@
 dependencies:
 - name: common
   repository: file://charts/common
-  version: 0.0.16
+  version: 0.0.17
```
oksana-grishchenko, line 4: _(the comment is not related to the selected lines, just wanted to have a thread somewhere)_

The PR looks good to me. Tried the fix in a FB, the upgrade works fine. 
The only thing that bothers is the failed required check in the CI, why does the error appear?

maxkondr: fixed

oksana-grishchenko (approved):

charts/everest/Makefile
```diff
@@ -42,3 +42,10 @@ CRD_DIR ?= config/crd
 CRD_VERSION ?= main
 crds-gen:
 	docker run --rm registry.k8s.io/kustomize/kustomize:v5.0.0 build $(EVEREST_REPO_URL)/$(CRD_DIR)?ref=$(CRD_VERSION) > crds/everest.yaml
+
+prepare-pr:
+	$(MAKE) deps
+	$(MAKE) docs-gen
+	EVEREST_REPO_URL=$(EVEREST_REPO_URL) \
+	CRD_DIR=$(CRD_DIR) \
```
mayankshah1607, line 50: Why do we do this? It looks redundant, but I'm sure there's another reason 🤔

maxkondr: because we need to pass it to another `$(MAKE)` call

PR #566, opened 2025-06-12
K8SPS-456: Allow setting custom environment variables

0xAX:
This commit adds support for defining `.Values.mysql.env` which will be propagated to the PerconaServerMySQL and as a result to the MySQL statefulset. That could be useful especially since Percona MySQL operator added support for BOOTSTRAP_READ_TIMEOUT.


0xAX:
@valmiranogueira Thank you!

PR #578, opened 2025-06-30
K8 spg 613

pooknull:


charts/pg-db/values.yaml
```diff
@@ -13,6 +13,32 @@ finalizers:
 #  - percona.com/delete-backups
 
 crVersion: 2.7.0
+#  initContainer:
```
valmiranogueira: Seems like initContainer has an indentation, is it expected? In CR initContainer is under spec key, not sure an indentation is necessary here.

charts/pg-db/values.yaml
```diff
@@ -340,6 +392,32 @@ backups:
   enabled: true
   trackLatestRestorableTime: true
   pgbackrest:
+#    initContainer:
```
valmiranogueira, line 395: Same question for indentation here

charts/pg-db/templates/_helpers.tpl
```diff
@@ -81,4 +81,21 @@ resources:
...
+{{/*
+Common initContainer template
+*/}}
+{{- define "pg-database.initContainer" -}}
+{{- if .initContainer }}
+initContainer:
+  {{- if .initContainer.image }}
+  image: {{ .initContainer.image }}
+  {{- end }}
+  {{- template "pg-database.resources" . }}
```
valmiranogueira: Not sure why, but resources doesn't seem to be generated in the template

charts/pg-db/values.yaml
valmiranogueira: This one is correct

jvpasinatto (approved):

PR #582, opened 2025-07-03
PMM-14063 openshift poc

BupycHuk:


charts/pmm/templates/NOTES.txt
```diff
@@ -18,7 +18,7 @@ Get the application URL:
 {{- else if contains "LoadBalancer" .Values.service.type }}
      NOTE: It may take a few minutes for the LoadBalancer IP to be available.
            You can watch the status of by running 'kubectl get --namespace {{ .Release.Namespace }} svc -w {{ .Values.service.name }}'
-  export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ .Values.service.name }} -o jsonpath="{.status.loadBalancer.ingress[0].ip}")
+  export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ .Values.service.name }} -o jsonpath="{.status.loadBalancer.ingress[0].hostname}")
```
ademidoff: I wonder if the hostname will always provide an IP address.

charts/pmm/values.yaml
```diff
@@ -11,7 +11,7 @@
 ## @param image.imagePullSecrets Global Docker registry secret names as an array
 ##
 image:
-  repository: percona/pmm-server
+  repository: perconalab/pmm-server
```
ademidoff: Is that not a typo or an experiment?

charts/pmm/values.yaml
```diff
@@ -79,14 +79,14 @@ secret:
...
...bnRyCnAzT2Y0UU5yRDVhSlNKK2U2QU9wWUFvU2EwWmQ3bFQ5TTZEZ1lLcFNlcUpnU3VGb1pJTEpReWRZbVp1N3RabFgKRUVzR1JoSnpyeGpZcDZkM3k3UUZZTFJFUFB6dVA5K2VBR2hKZVZtM0lPS3NaS2JIYklaY29zSFJKeXRIUDlNawpkdWRFUG1rOUFZU1I0R1NoZUJiYnp2ajZpUUtCZ1FEZ2lUYlV4TWhnOW51T3hWYkhzUmVId24zRDlQOVFiUVQyCkdoRXhjWTZYSE9SZXNpQTFyY1dsK0RpNGV4d1NKMkZDNjM2Y2NzcjFzd0N5Y0Y5SVd1UGF3Z29hWXdpSURIUUwKRmtXTitZUWlKTWZzMVRQVGhTVkpyTVl4aHd1UkV3bzY3MjdZUFQ2YWU3Y0pMRTJyZUxLUExybmN6UEdyK0ZSMApWU3dTZmYyb3JRS0JnUURodjErUld2YUxUVGhMWXA2MlJubEpYQ3hRTnRYOHFMODFuK3djR3doc240ZFJQNTJPCm5FdmhSdHZwcGtCTFNDVEVUMysxVWszRU1iTWxGalp2dFlhTUNFc2pUWWRHL25TdzViNFdtVVNoTmRJcytHL1kKWE1SdDdGaElwU0tEMGVUcUpWR25oK1NscUhXbENIVnhrb1k1T0lWTmgwQ3BtNDI4V3J4a1JTZDRDUUtCZ0daNgpXWnNjUVBQNEt3WkxUTU1STVZNVjBqNllWOXFHRjBpWE54ZWc0MXBXcG1nTjFpT0pydmFENXFkVUVNaThUbWl6CmdISzQ5YWJUenRLaDlPcmFDTHB6NnhhRjhvUGRZeTBZT2lDRksraEZielhKVE1FU09LVjZoRzVnVzBnWUw0Z2MKZFRPZUtDOXJuV3REWENPUmo3Vkh0eTV1SHpMRXdRN2pQSjBvQVpoMUFvR0JBS0FFbEw1K3lJTktxQ09CNnBsSQpDbUp6aHNraldEQXNncmx6dTlsZnBVNFpDSTFuOWdjMDR5N2VXdTFkYmNmTjdLcFJJSVY0bmZZM205WnNYVEwwCkFIYlRKdFR2SkhkN2JCMFV3NEFIazdrMllEUWQvY1FIZ21wTGdrZ1IvdFBSZ3ErNTMvSGJoQ1V1SUVydGRwdFUKenRhdzdpTVRhOUZEVUpGRkFnME9uK1FaCi0tLS0tRU5EIFBSSVZBVEUgS0VZLS0tLS0K
```
ademidoff: Can/should we really commit this private key?

charts/pmm/values.yaml
```diff
@@ -79,14 +79,14 @@ secret:
...
 ##   please provide your own signed ssl certificates like this in base 64 format:
-## certs:
-  ## name: pmm-certs
-  ## files:
-  ##  certificate.crt:
-  ##  certificate.key:
-  ##  ca-certs.pem:
-  ##  dhparam.pem:
-  ##  certificate.conf:
+#certs:
```
JiriCtvrtka: Isn't it better to let the PMM generate it? Or there is reason why to use specific one?

charts/pmm/templates/_helpers.tpl
```diff
@@ -84,3 +84,15 @@ checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sh
...
+Create password if it does not exist or reuse existing one.
+*/}}
+{{- define "pmm.password" -}}
+{{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.secret.name -}}
+{{- if and $secret $secret.data.PMM_ADMIN_PASSWORD -}}
+{{ $secret.data.PMM_ADMIN_PASSWORD }}
+{{- else -}}
+{{ .Values.secret.pmm_password | default (randAscii 16) | b64enc }}
+{{- end -}}
+{{- end -}}
```
JiriCtvrtka: small one: missing empty line

charts/pmm/values.yaml
BupycHuk: will revert it

charts/pmm/values.yaml
BupycHuk: will revert it

charts/pmm/templates/NOTES.txt
BupycHuk: Hi, in my case it was providing hostname

JiriCtvrtka (dismissed):

charts/pmm/templates/NOTES.txt
ademidoff: My point is: we are exporting a variable called `SERVICE_IP`, which is supposed to contain an IP address, but in reality we are storing a hostname to it, which does not necessarily have to be an IP address. Maybe consider changing the name then?

charts/pmm/values.yaml
```diff
@@ -232,6 +232,11 @@ podSecurityContext:
   runAsGroup: 1000
   fsGroupChangePolicy: OnRootMismatch
   fsGroup: 1000
+# to make it work with openshift comment everything above and uncomment lines below
```
ademidoff, line 277: Would it work for Openshift and non-OpenShift if we uncomment the lines below and comment out the lines above?

charts/pmm/templates/NOTES.txt
BupycHuk: ok, let me revert this change as it's not related to the ticket. We need a proper implementation of this.

PR #584, opened 2025-07-03
[pg-db] set spec.backups.pgbackrest.manual only if defined

bastiaanb:
This fixes an issue when deploying a pg-db chart with ArgoCD:
The Percona operator modifies `spec.backups.pgbackrest.manual.options` and `spec.backups.pgbackrest.manual.repoName` during backups  (see K8SPG-816).
If autosync is enabled, ArgoCD would revert those changes, causing the backup to never complete. 

To solve this, set ServerSideApply=true in ArgoCD and set `spec.backups.pgbackrest.manual` to nil.
ArgoCD will then not become field-manager for that field and ignore changes in that field by Percona operator.


bastiaanb:
Signed CLA.

jvpasinatto:
@bastiaanb Thanks for the contribution!

bastiaanb:
@jvpasinatto thanks for merging and releasing!

PR #590, opened 2025-07-21
K8SPG-833: add `env` and `envFrom` fields

pooknull:
https://perconadev.atlassian.net/browse/K8SPG-833


jvpasinatto (changes requested):
Please change base branch to `pg-2.8.0`

PR #604, opened 2025-08-06
PSMDB operator - support revisionhistorylimit

DrFaust92:


charts/psmdb-operator/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: "1.20.1"
 description: A Helm chart for deploying the Percona Operator for MongoDB
 name: psmdb-operator
 home: https://docs.percona.com/percona-operator-for-mongodb/
-version: 1.20.1
+version: 1.21.0
```
hors: @DrFaust92 we did not release 1.21.0 It is under QA now. Please use 1.20.2

DrFaust92: thanks, done!

charts/psmdb-operator/templates/deployment.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
   {{- end }}
 spec:
   replicas: {{ .Values.replicaCount }}
+  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
```
valmiranogueira: Have '| default 10' would be good I think (maybe someone forgets or removes the item from values by mistake?)

DrFaust92: if someone removes this it will take it from the chart default values.

valmiranogueira: Not in case a --values flag is used during installation/upgrade

valmiranogueira: Please add an if for this value or use | default, to avoid any unset or old values.yaml (same approach is also used for new values introduced in all operators)

charts/psmdb-operator/values.yaml
```diff
@@ -4,6 +4,8 @@
 
 replicaCount: 1
 
+revisionHistoryLimit: null
```
valmiranogueira: I know it makes no difference, but as standard we leave in values file the default (as 10).

In addition to that please add the parameter to README.

Thank you a lot!

DrFaust92: [valmiranogueira](https://github.com/valmiranogueira), done.

PR #617, opened 2025-08-21
EVEREST-2177 Add default lbc, initContainer and lbc cleanup

oksana-grishchenko:
EVEREST-2177 

Add default lbc, initContainer and lbc cleanup

Related PRs:
- https://github.com/percona/everest/pull/1590
- https://github.com/percona/everest-operator/pull/861

### DIFF
--- a/charts/everest/templates/everest-operator/deployment.yaml
+++ b/charts/everest/templates/everest-operator/deployment.yaml
@@ -19,6 +19,7 @@ spec:
       labels:
         app: everest-operator
     spec:
+      {{- if .Values.operator.init }}
       initContainers:
       - name: crs-migration
         env:
@@ -28,6 +29,7 @@ spec:
         command:
         - /migrator
         imagePullPolicy: IfNotPresent
+      {{- end }}
       containers:
       - args:
         {{- if .Values.operator.enableLeaderElection }}
--- a/charts/everest/templates/everest-operator/deployment.yaml
+++ b/charts/everest/templates/everest-operator/deployment.yaml
@@ -19,6 +19,17 @@ spec:
       labels:
         app: everest-operator
     spec:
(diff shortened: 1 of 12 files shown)


oksana-grishchenko:
Not sure if it is expected that the CI fails, figuring it out. It timed-out waiting for the everest-operator availability, which could be because the operator changes were not merged.

charts/everest/templates/everest-operator/deployment.yaml
```diff
@@ -19,6 +19,17 @@ spec:
...
+      - name: crs-migration
+        env:
+        - name: EVEREST_VERSION
+          value: {{ .Chart.Version }}
+        - name: SYSTEM_NAMESPACE
+          value: {{ include  "everest.namespace" . }}
+        image: {{ .Values.operator.image }}:{{ .Chart.Version }}
+        command:
+        - /migrator
+        imagePullPolicy: IfNotPresent
```
mayankshah1607: Shall we put this behind a variable which is `true` by default? 
```suggestion
      {{- if .Values.operator.init }}
      initContainers:
      - name: crs-migration
        env:
        - name: EVEREST_VERSION
          value: {{ .Chart.Version }}
        - name: SYSTEM_NAMESPACE
          value: {{ include  "everest.namespace" . }}
        image: {{ .Values.operator.image }}:{{ .Chart.Version }}
        command:
        - /migrator
        imagePullPolicy: IfNotPresent
      {{- end }}
```

That way in https://github.com/percona/everest/pull/1590 we don't have to do any YAML patching, just set this value to `false`

oksana-grishchenko: updated, thanks!

mayankshah1607 (approved):

PR #636, opened 2025-09-15
EVEREST-2303 | Move from bitnami/kubectl image

mayankshah1607:
### Problem

Everest Helm chart uses the `bitnami/kubectl` Docker image to execute some chart hooks. However, the `bitnami` registry is planned for deletion - https://github.com/bitnami/charts/issues/35164

### Solution

- move to `bitnamilegacy/kubectl` temporarily. We might eventually need to build are own image, but this will buy us some extra time
- provide a way to override the image used in the chart hook so that we are not so tightly dependent on an external vendor, and have the flexibility to use alternate images in such situations

### DIFF
--- a/charts/everest/values.yaml
+++ b/charts/everest/values.yaml
@@ -1,7 +1,7 @@
 # -- Configuration for Helm chart hooks.
 hooks:
   # -- Default image to use for the Helm chart hooks job.
-  # @default -- `registry.k8s.io/kubectl`
+  # @default -- `bitnamilegacy/kubectl`
   image: bitnamilegacy/kubectl
   pspCleanup: {}
     # -- Image to use for the PSP cleanup job.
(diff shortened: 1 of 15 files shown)


charts/everest/values.yaml
```diff
@@ -1,3 +1,18 @@
+# -- Configuration for Helm chart hooks.
+hooks:
+  # -- Default image to use for the Helm chart hooks job.
+  # @default -- `bitnamilegacy/kubectl`
+  image: bitnamilegacy/kubectl
```
mayankshah1607, line 5: We will have to move away from this as well eventually, but this will at least buy us some time. Alternatives considered were:
- `registry.k8s.io/kubectl` -- but it lacks `bin/sh` which is needed for our current scripts
- `alpine/k8s` -- contains lots of k8s related tools (helm, kustomize, eksctl, kubectl, etc.) so the image is huge (1.5G)

We probably might need to build our own image

PR #641, opened 2025-09-17
PMM-14308 Fix indentation for namespace inclusion

BupycHuk:
PMM-14308


ademidoff (dismissed):

ademidoff:
@BupycHuk 

It seems like we also need to bump the version.

> pmm => (version: "1.4.11", path: "charts/pmm") > chart version not ok. Needs a version bump!

PR #667, opened 2025-10-10
PMM-14395 - Update PMM HA chart to use external repository for Victor…

theTibi:
…iaMetrics

- Changed the repository for the VictoriaMetrics cluster dependency in Chart.yaml and Chart.lock from a local file path to the official external repository URL.
- Updated the Chart.lock file to reflect the new digest and generated timestamp.
- Enhanced documentation in values.yaml regarding ClickHouse user management and added instructions for retrieving VictoriaMetrics credentials from Kubernetes secrets.

### DIFF
--- a/charts/pmm/Chart.yaml
+++ b/charts/pmm/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: pmm
 description: A Helm chart for Percona Monitoring and Management (PMM)
 type: application
-version: 1.4.14
-appVersion: "3.4.0"
+version: 1.4.13
+appVersion: "3.4.1"
 home: https://github.com/percona/pmm
 maintainers:
   - name: tplavcic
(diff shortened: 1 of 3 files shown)


charts/pmm-ha/values.yaml
```diff
@@ -103,9 +103,14 @@ secret:
...
-  ## @param secret.clickhouse_user ClickHouse username. If not set, defaults to "clickhouse_pmm"
+  ## @param secret.clickhouse_user ClickHouse user. If not set, defaults to "clickhouse_pmm"
+  ## When secret.create: true and this is empty, "clickhouse_pmm" is used.
+  ## If creating the secret manually (secret.create: false), this field is ignored.
+  ##
+  ## To get username after install:
+  ##   kubectl get secret pmm-secret -o jsonpath='{.data.PMM_CLICKHOUSE_USER}' | base64 --decode
   ## E.g.
-  ## clickhouse_user: clickhouse_pmm
+  ## clickhouse_user: my_clickhouse_user
```
BupycHuk: according to text above it should default to clickhouse_pmm

PR #668, opened 2025-10-13
PMM-14396 - Update PMM HA chart to integrate Percona PostgreSQL Operator

theTibi:
- Replaced local pg-db chart references with the Percona Helm repository in Chart.yaml and Chart.lock for better dependency management.
- Updated init.sql to remove hardcoded database and user creation, now relying on Helm to manage user secrets.
- Enhanced values.yaml to include user configuration for PostgreSQL, allowing for automated user and database setup.
- Removed obsolete pg-db chart files and templates to streamline the project structure and reduce redundancy.

### DIFF
--- a/charts/pmm-ha/templates/hooks/pre-delete.yaml
+++ b/charts/pmm-ha/templates/hooks/pre-delete.yaml
@@ -22,40 +22,74 @@ spec:
           - |
             set -euo pipefail
             
-            # Function to clean up custom resources
-            cleanup_resource() {
+            NAMESPACE="{{ .Release.Namespace }}"
+            RELEASE_NAME="{{ .Release.Name }}"
+            PG_CLUSTER_NAME="${RELEASE_NAME}-pg-db"
+            LABEL="postgres-operator.crunchydata.com/cluster=$PG_CLUSTER_NAME"
+            
+            echo "=========================================="
+            echo "Pre-delete: Cleaning up PostgreSQL resources"
+            echo "Namespace: $NAMESPACE"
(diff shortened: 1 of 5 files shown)


charts/pmm-ha/templates/hooks/pre-delete.yaml
```diff
@@ -0,0 +1,66 @@
...
+    "helm.sh/hook-weight": "-1"
+    "helm.sh/hook-delete-policy": "before-hook-creation,hook-succeeded"
+spec:
+  template:
+    spec:
+      serviceAccountName: {{ include "pmm.serviceAccountName" . }}
+      restartPolicy: Never
+      containers:
+      - name: check-clusters
+        image: "ubuntu:22.04"
```
ademidoff:
```suggestion
        image: "alpine/kubectl:latest"
```
That would avoid installing `kubectl`. FYI: we already use it in our chart.

charts/pmm-ha/templates/pg-secret-init-job.yaml
```diff
@@ -17,27 +33,43 @@ rules:
...
 metadata:
-  name: secret-init-job
+  name: {{ .Release.Name }}-secret-init-job
   namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "pmm.labels" . | nindent 4 }}
+  annotations:
+    meta.helm.sh/release-name: {{ .Release.Name }}
+    meta.helm.sh/release-namespace: {{ .Release.Namespace }}
+    "helm.sh/hook": pre-install,pre-upgrade
```
ademidoff: I guess we should not generate the secrets on each upgrade, it's a one-time job.

PR #687, opened 2025-10-31
PMM-14429 Pass the encryption key to PMM

ademidoff:


charts/pmm-ha/templates/statefulset.yaml
```diff
@@ -40,6 +40,20 @@ spec:
...
+          command:
+            - /bin/sh
+            - -c
+            - |
+              echo "Waiting for secret to be available..."
+              until kubectl get secret pg-encryption-key; do
+                echo "Secret not found, retrying in 5s..."
+                sleep 5
+              done
+              echo "Secret found, proceeding..."
```
BupycHuk, line 56: I think it can be handled by using pre-install hook for the job.

charts/pmm-ha/templates/pg-secret-rotation-cronjob.yaml
```diff
@@ -0,0 +1,77 @@
...
+kind: CronJob
+metadata:
+  name: secret-rotation
+  namespace: {{ .Release.Namespace }}
+spec:
+  # Rotate every 30 days at 2 AM (adjust as needed)
+  schedule: "0 2 1 * *"
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 3
+  jobTemplate:
```
BupycHuk: Rotation of the encryption key should include re-encryption of existing data otherwise it will lead to failure when PMM will use new encryption key to decrypt old data.

ademidoff: Agreed, thanks. I will remove it from here and address it in a separate ticket.

charts/pmm-ha/templates/statefulset.yaml
ademidoff: I've learned some time ago that hooks should be used with caution and as a last resort.

charts/pmm-ha/templates/pg-secret-init-job.yaml
```diff
@@ -0,0 +1,95 @@
...
+          # Generate the secret using pmm-encryption-rotation and write to shared volume
+          /usr/sbin/pmm-encryption-rotation --generate-key > /shared/secret-value
+          
+          echo "Key generated successfully"
+        volumeMounts:
+        - name: shared-data
+          mountPath: /shared
+      containers:
+      - name: kubectl-applier
+        image: bitnami/kubectl:latest
```
Lappihuan: Bitnami should probably not be reintroduced

ademidoff: Do you have a good alternative in mind?

PR #688, opened 2025-11-05
K8SPXC-1332: add caBundle

mayankshah1607:


charts/pxc-db/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: 1.18.0
 description: A Helm chart for installing Percona XtraDB Cluster Databases using the PXC Operator.
 name: pxc-db
 home: https://www.percona.com/doc/kubernetes-operator-for-pxc/kubernetes.html
-version: 1.18.0
+version: 1.18.1
```
mayankshah1607: What should the updated chart version be? I assume this CA Bundle feature will be included in 1.19.0? So should I change appVersion and version to be 1.19.0?

gkech: We have already a base branch with these changes, please change your PR here to use as base: https://github.com/percona/percona-helm-charts/tree/release-pxco-1.19.0

mayankshah1607: done, thanks!

mayankshah1607: @gkech but the chart linter still complains, is it expected?

PR #692, opened 2025-11-08
Fix accessModes handling in cluster.yaml

oliverisaac:
When accessMode is not defined you get then accessModes: is not included:

```
storageClass: example
- ReadWriteMany
resources: ...
```

This change fixes it so that accessModes: is outside the if statement


gkech:
thanks for the contribution, prepared a ticket for this: https://perconadev.atlassian.net/browse/K8SPG-1038

gkech:
@oliverisaac can you please sign the CLA?

PR #702, opened 2025-11-20
EVEREST-2306 PMM3 separate variable

oksana-grishchenko:
In order to prevent the existing PMM installation made by setting `pmm.enabled=true` from being corrupted we introduce a new variable `pmm3`, which should be used the same way the `pmm` variable was used but it will install PMM3 instead of PMM2. 

**Highlights:**
`pmm.enabled` and `pmm3.enabled` can't be set to true simultaniously. 

- if we just try to install  Everest with both `pmm.enabled=true` `pmm3.enabled=true`, it results:

```
Installing Percona Everest version 1.10000.0-rc20251120124504...
Error: INSTALLATION FAILED: execution error at (everest/templates/everest-operator/deployment.yaml:1:3): Only one of pmm.enabled or pmm3.enabled may be true. They are mutually exclusive.
```

- if there was an installation with `pmm.enabled=true` and we try to upgrade Everest with   `--reset-then-reuse-values
  --set pmm3.enabled=true`, it results:

```
Deploying Percona Version Service...
Upgrading Percona Everest CRDs...
Release "everest-crds" does not exist. Installing it now.
NAME: everest-crds
LAST DEPLOYED: Thu Nov 20 18:21:42 2025
NAMESPACE: everest-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
Upgrading Percona Everest to version 1.10000.0-rc20251120124504...
Error: UPGRADE FAILED: execution error at (everest/templates/everest-operator/deployment.yaml:1:3): Only one of pmm.enabled or pmm3.enabled may be true. They are mutually exclusive.
```
- if there was an installation with `pmm.enabled=true` and we try to upgrade Everest with   `--reset-then-reuse-values` then the upgrade is successful, no restarts on the existing pmm pods observed

### DIFF
--- a/charts/everest/README.md
+++ b/charts/everest/README.md
@@ -205,7 +205,9 @@ The following table shows the configurable parameters of the Percona Everest cha
 | operator.webhook.certs | object | `{"ca.crt":"","tls.crt":"","tls.key":""}` | Certificates to use for the webhook server. The values must be base64 encoded. If unset, uses self-signed certificates. |
 | operator.webhook.preserveTLSCerts | bool |  | If set to true, preserves existing TLS Certificate Secrets during upgrades. This setting is ignored if certificates are explicitly provided in operator.webhook.certs, in which case the specified certificates are used instead. This setting has no effect during installation. |
 | pmm | object | `{"enabled":false,"nameOverride":"pmm"}` | PMM settings. |
-| pmm.enabled | bool | `false` | If set, deploys PMM in the release namespace. |
+| pmm.enabled | bool | `false` | If set, deploys PMM2 in the release namespace. |
+| pmm3.enabled | bool | `false` | If set, deploys PMM3 in the release namespace. |
+| pmm3.nameOverride | string | `"pmm3"` |  |
 | server.apiRequestsRateLimit | int | `100` | Set the allowed number of requests per second. |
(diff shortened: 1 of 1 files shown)


charts/everest/values.yaml
```diff
@@ -313,12 +299,15 @@ monitoring:
...
 # -- PMM settings.
 pmm:
-  # -- If set, deploys PMM in the release namespace.
+  # -- If set, deploys PMM2 in the release namespace.
   enabled: false
   nameOverride: pmm
+pmm3:
+  # -- If set, deploys PMM3 in the release namespace.
+  enabled: false
+  nameOverride: pmm3
```
oksana-grishchenko, line 310: Highlighting the only significant change in this file. 
Other changes (indents) appears automatically when I run `make release-dev`

charts/everest/Chart.yaml
```diff
@@ -36,6 +36,11 @@ dependencies:
   - name: pmm
     repository: "https://percona.github.io/percona-helm-charts"
     condition: "pmm.enabled"
+    version: 1.3.*
+  - name: pmm
+    alias: pmm3
```
oksana-grishchenko, line 41: Using an alias to be able to have two versions of the same dependency

PR #703, opened 2025-11-22
Everest - make PMM3 a subchart to avoid conflicts with PMM2

recharte:


charts/everest/values.yaml
```diff
@@ -307,7 +307,9 @@ pmm:
 pmm3:
   # -- If set, deploys PMM3 in the release namespace.
   enabled: false
-  nameOverride: pmm3
+  # -- PMM configuration. All PMM chart values go under this key.
+  pmm:
+    nameOverride: pmm3
```
oksana-grishchenko, line 312: Am I understand it correctly that now to set any values like `pmm.secret.pmm_password=admin` we should use `pmm3.pmm.secret.pmm_password=admin`? We probably should mention it in the docs then

oksana-grishchenko (approved):

charts/everest/values.yaml
recharte: Yes, I couldn't find a way of passing it as we had previously

PR #711, opened 2025-11-26
K8SPXC-1525 deprecate probes

nmarukovich:


nmarukovich:
> We need to remove them from cluster.yaml as well

no! we don't need. We delete only proxysql because it doesn't work, the rest we should keep until we delete it completely.

jvpasinatto (approved):

PR #724, opened 2025-12-09
K8SPXC-735: Add scheduler fields

egegunes:


charts/pxc-db/values.yaml
```diff
@@ -221,7 +221,7 @@ pxc:
   # serviceAccountName: percona-xtradb-cluster-operator-workload
 
 haproxy:
-  enabled: true
+  enabled: false
```
jvpasinatto: pls revert this and `proxysql.enabled`

egegunes: updated

jvpasinatto (approved):

PR #757, opened 2026-01-09
Fix instance metadata

aleixab:
Fixes an issue where metadata (labels/annotations) defined in instances were ignored during templating.

### DIFF
--- a/charts/pg-db/templates/cluster.yaml
+++ b/charts/pg-db/templates/cluster.yaml
@@ -238,7 +238,7 @@ spec:
   instances:
   {{- range $instance := .Values.instances }}
     - name: {{ $instance.name }}
-      {{- if or ($instance.metadata) }}
+      {{- if and $instance.metadata (or $instance.metadata.annotations $instance.metadata.labels) }}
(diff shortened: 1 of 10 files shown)


charts/pg-db/templates/cluster.yaml
```diff
@@ -202,6 +202,17 @@ spec:
   instances:
   {{- range $instance := .Values.instances }}
     - name: {{ $instance.name }}
+      {{- if or ($instance.metadata) }}
```
valmiranogueira: why or?

jvpasinatto: also, please make it render metadata only if annotations or labels are non empty

gkech:
@aleixab can you please review comments?

charts/pg-db/templates/cluster.yaml
aleixab: Fixed.

aleixab:
Applied comments

jvpasinatto (approved):

PR #760, opened 2026-01-09
PMM-14697 - HAProxy configuration for dynamic DNS discovery

theTibi:
- Introduced maxReplicas parameter in values.yaml to set the maximum number of PMM replicas for HAProxy server-template.
- Updated haproxy-configmap.yaml to utilize server-template for dynamic DNS-based discovery, allowing automatic pod discovery during scaling without requiring HAProxy restarts.
- Enhanced comments for clarity on the configuration and behavior of HAProxy with respect to PMM pods.

### DIFF
--- a/.github/workflows/pmm-ha-pr-checks.yaml
+++ b/.github/workflows/pmm-ha-pr-checks.yaml
@@ -92,6 +92,12 @@ jobs:
           
           echo "All operators are ready!"
 
+      # - name: Setup tmate session on failure
+      #   uses: mxschmitt/action-tmate@v3
+      #   timeout-minutes: 60
+      #   with:
+      #     limit-access-to-actor: true 
+
       - name: Run chart-testing (install pmm-ha)
         if: steps.list-changed.outputs.changed == 'true'
         run: |
--- a/charts/pmm-ha/README.md
+++ b/charts/pmm-ha/README.md
@@ -464,7 +464,6 @@ The PMM HA deployment includes several HA-specific environment variables:
 ```yaml
 pmmEnv:
   PMM_HA_ENABLE: "1"               # Enable HA clustering
-  PMM_HA_BOOTSTRAP: "1"            # Bootstrap HA cluster
   PMM_HA_GOSSIP_PORT: "9096"       # Gossip protocol port
   PMM_HA_RAFT_PORT: "9097"         # Raft consensus port
   PMM_HA_GRAFANA_GOSSIP_PORT: "9094"  # Grafana gossip port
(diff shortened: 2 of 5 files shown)


charts/pmm-ha/templates/haproxy-configmap.yaml
```diff
@@ -56,10 +56,8 @@ data:
...
-        {{- if eq $i 0 }}
-        server {{ $.Release.Name }}-{{ $i }}-https {{ $.Release.Name }}-{{ $i }}.{{ $.Values.service.name | default "monitoring-service" }}.{{ $.Release.Namespace }}.svc.cluster.local:8443 check ssl verify none resolvers k8s init-addr last,libc,none
-        {{- else }}
-        server {{ $.Release.Name }}-{{ $i }}-https {{ $.Release.Name }}-{{ $i }}.{{ $.Values.service.name | default "monitoring-service" }}.{{ $.Release.Namespace }}.svc.cluster.local:8443 check ssl verify none resolvers k8s init-addr last,libc,none
-        {{- end }}
-        {{- end }}
+        # Use server-template for dynamic DNS-based discovery of PMM pods
+        # This automatically discovers pods when scaling up/down without requiring HAProxy restart
+        # HAProxy will re-resolve DNS based on 'hold valid' setting in the resolver (10s)
+        server-template pmm 1-{{ $.Values.maxReplicas | default 10 }} {{ $.Values.service.name | default "monitoring-service" }}.{{ $.Release.Namespace }}.svc.cluster.local:8443 check ssl verify none resolvers k8s init-addr last,libc,none
```
ademidoff, line 62: Question: Don't we have to use `{{ $.Release.Name }}` instead of `pmm` in here?

theTibi: These are not pod names this is just a prefix for Haproxy slots. Haproxy will keep 10 slots, then queries DNS for the headless service (monitoring-service.namespace.svc.cluster.local) and will use those IPs  of the pods and assign them to a slot. The actual pod name does not matter for haproxy.  if we want we can use something like this:
` 
server-template {{ $.Release.Name }}- 1-{{ $.Values.maxReplicas | default 10 }} {{ $.Values.service.name | default "monitoring-service" }}.{{ $.Release.Namespace }}.svc.cluster.local:8443 check ssl verify none resolvers k8s init-addr last,libc,none
`

That will make hapoxy stats maybe more clearer but again these names will not correspond with the real pods, these are just reserved slots and haproxy will assign the pods randomly to the slots..

PR #771, opened 2026-01-19
Fix cluster accessModes template

### LINKED ISSUE #766: pg-db helm chart error, issue with cluster template
Look at the cluster template: https://github.com/percona/percona-helm-charts/blob/main/charts/pg-db/templates/cluster.yaml
and the attached image

Image

if we define dataVolumeClaimSpec without specify the access mode this would render

```yaml
dataVolumeClaimSpec:
  - ReadWriteOnce
```

instead of the expectation is

```yaml
dataVolumeClaimSpec:
  accessModes:
    - ReadWriteOnce
```

that will cause of failing

aleixab:
Fixes: #766

### DIFF
--- a/charts/pg-db/Chart.yaml
+++ b/charts/pg-db/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: pg-db
 description: 'A Helm chart to deploy the PostgreSQL database with the Percona Operator for PostgreSQL'
 type: application
-version: 2.9.0
+version: 2.9.1
 appVersion: 2.9.0
 home: https://docs.percona.com/percona-operator-for-postgresql/2.0/
 maintainers:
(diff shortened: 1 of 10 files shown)


charts/pg-db/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: pg-db
 description: 'A Helm chart to deploy the PostgreSQL database with the Percona Operator for PostgreSQL'
 type: application
-version: 2.8.2
+version: 2.8.3
```
valmiranogueira: Please update the README with newer chart version like:
`helm install my-db percona/pg-db --version 2.8.3 --namespace my-namespace`

jvpasinatto (changes requested):
Please apply the same fix for `instances.walVolumeClaimSpec` and `backups.pgbackrest.repos[].volume.volumeClaimSpec `

gkech:
@aleixab can you please have a look or the comments? otherwise we will take over :)

charts/pg-db/Chart.yaml
aleixab: Fixed

aleixab:
Applied comments

jvpasinatto (approved):

PR #782, opened 2026-01-28
Add `clusterServiceDNSSuffix` field

pooknull:
https://perconadev.atlassian.net/browse/K8SPG-694

### DIFF
--- a/charts/pg-db/values.yaml
+++ b/charts/pg-db/values.yaml
@@ -48,7 +48,7 @@ repository: docker.io/percona/percona-distribution-postgresql
 image: docker.io/percona/percona-distribution-postgresql:17.7-2
 imagePullPolicy: Always
 postgresVersion: 17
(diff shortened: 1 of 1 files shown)


charts/pg-db/values.yaml
```diff
@@ -48,6 +48,7 @@ repository: docker.io/percona/percona-distribution-postgresql
 image: docker.io/percona/percona-distribution-postgresql:17.7-2
 imagePullPolicy: Always
 postgresVersion: 17
+clusterServiceDNSSuffix: cluster.local
```
valmiranogueira: It shouldn't have a default value, but follow the CR which is a comment

pooknull: https://github.com/percona/percona-helm-charts/pull/782/commits/e78fb685a90da14e7cc68669feccc9993deb3289

PR #786, opened 2026-02-10
PMM-14821 - Add Helm repositories and build chart dependencies in release workflow

theTibi:
pmm-ha-dependencies chart has a few dependencies and because of that the release job is now failing on github.

`Looking up latest tag...
Discovering changed charts since 'everest-1.13.0'...
Installing chart-releaser on /opt/hostedtoolcache/cr/v1.6.1/x86_64...
Adding cr directory to PATH...
Packaging chart 'charts/pmm-ha-dependencies'...
Error: no repository definition for https://victoriametrics.github.io/helm-charts/, https://helm.altinity.com,/ https://percona.github.io/percona-helm-charts/`

I am adding this dependencies to the `release.yaml`. That should not impact the other charts, because charts that don't have external dependencies simply won't use those repos. And for charts that do have dependencies, having the repos registered just means Helm can resolve them if needed.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -31,6 +31,17 @@ jobs:
         with:
           version: v3.15.4
 
+      - name: Add Helm repositories
+        run: |
+          helm repo add victoriametrics https://victoriametrics.github.io/helm-charts/
+          helm repo add altinity https://helm.altinity.com
+          helm repo add percona https://percona.github.io/percona-helm-charts/
+          helm repo update
+
+      - name: Build chart dependencies
+        run: |
+          helm dependency build charts/pmm-ha-dependencies
+
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.6.0
         env:


jvpasinatto (approved):

.github/workflows/release.yaml
```diff
@@ -31,6 +31,17 @@ jobs:
         with:
           version: v3.15.4
 
+      - name: Add Helm repositories
```
valmiranogueira, line 34: I'm wondering if it's possible to add condition here to run only when changes were made to charts/pmm-ha-dependencies (or any other related chart)

theTibi: @valmiranogueira something like this?

``` 
      - name: Detect charts with dependencies
        id: detect-deps
        run: |
          charts_with_deps=""
          if git diff HEAD~1 --name-only | grep -q '^charts/pmm-ha-dependencies/'; then
            charts_with_deps="charts/pmm-ha-dependencies $charts_with_deps"
          fi
          echo "charts=$charts_with_deps" >> $GITHUB_OUTPUT
          echo "has_deps=$([ -n "$charts_with_deps" ] && echo true || echo false)" >> $GITHUB_OUTPUT

      - name: Add Helm repositories
        if: steps.detect-deps.outputs.has_deps == 'true'
        run: |
          helm repo add victoriametrics https://victoriametrics.github.io/helm-charts/
          helm repo add altinity https://helm.altinity.com
          helm repo add percona https://percona.github.io/percona-helm-charts/
          helm repo update

      - name: Build chart dependencies
        if: steps.detect-deps.outputs.has_deps == 'true'
        run: |
          for chart in ${{ steps.detect-deps.outputs.charts }}; do
            echo "Building dependencies for $chart..."
            helm dependency build "$chart"
          done```

valmiranogueira: Yeah, I was just thinking that if helm dependency build could make the charts/pmm-ha-dependencies to be always released on every push to main. But probably not, cause the chart version doesn't change all the time

valmiranogueira (approved):

PR #791, opened 2026-02-16
K8SPG-758: allow setting PPROF_BIND_ADDRESS

mayankshah1607:


charts/pg-operator/values.yaml
```diff
@@ -19,6 +19,11 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
+
+# The TCP address that the controller should bind to for serving pprof.
+# Setting "" or "0" will disable pprof serving.
+# pprofBindAddress: ""
```
valmiranogueira: This is adding conditionally, but in the operator default value is 0

mayankshah1607: https://github.com/percona/percona-helm-charts/pull/791/changes/61549de2f0ae56c4659979f523ae113a81d5a079

charts/pg-operator/values.yaml
```diff
@@ -19,6 +19,8 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
+
```
valmiranogueira: Are these extra lines intended? Lint is failing

mayankshah1607: done

PR #796, opened 2026-02-17
Only Output Roles when Orchestrator enabled

utdrmac:


charts/ps-db/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: "1.0.0"
+appVersion: "1.0.1"
```
valmiranogueira: App version remains 1.0.0

charts/ps-db/Chart.yaml
```diff
@@ -1,9 +1,9 @@
...
 description: A Helm chart for installing Percona Server Databases using the PS Operator.
 name: ps-db
 home: https://www.percona.com/doc/kubernetes-operator-for-mysql/ps
-version: 1.0.0
+version: 1.0.1
```
valmiranogueira, line 6: When changing chart version, please update README with the newer version

utdrmac: There is nothing in the README about version history. What exactly do you want me to update?

valmiranogueira: helm install my-db percona/ps-db --version `new-version` --namespace my-namespace

utdrmac: I see now. Thanks.

valmiranogueira:
Hi @utdrmac I was testing this PR, and actually these access are now managed by operator, and they shouldn't be part of Helm. In this case these templates should be actually removed.

PR #810, opened 2026-03-03
K8SPG-911: Add pg_tde and change extensions structure

egegunes:


jvpasinatto (commented):
can you remove pg_tde for now?

PR #811, opened 2026-03-05
Fixed indentation for proxysql image pull secrets

JustInVTime:
When using pull secrets for the proxysql image, the indentation is wrong. This fixes it


charts/pxc-db/templates/cluster.yaml
```diff
@@ -387,7 +387,7 @@ spec:
     {{- end }}
     {{- if $proxysql.imagePullSecrets }}
     imagePullSecrets:
-    {{- $proxysql.imagePullSecrets | toYaml | indent 6 }}
+{{ $proxysql.imagePullSecrets | toYaml | indent 6 }}
```
valmiranogueira, line 390: In order to not rely on how it's indented on template functions, the optimal solution would be:
`{{- $proxysql.imagePullSecrets | toYaml | nindent 6 }}`
And keep template indentation for readability

charts/pxc-db/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: 1.19.0
+appVersion: 1.19.1
```
valmiranogueira: Actually lint failed, cause chart version needs bump, app versions stay the same

JustInVTime: My bad, should be fixed now

valmiranogueira: But please keep 1.19.0 as appVersion, since this is the version of released operator

JustInVTime: done

PR #819, opened 2026-03-20
K8SPXC-1820: add `helm-unittest` tests for pxc operator

pooknull:
https://perconadev.atlassian.net/browse/K8SPXC-1820

This PR creates a basic unit tests for both `pxc-operator` and `pxc-db` charts using [`helm-unittest`](https://github.com/helm-unittest/helm-unittest).

It also adds `Makefile` targets to install the plugin and run the tests with `make test`.


.github/workflows/test.yaml
```diff
@@ -7,6 +7,30 @@ on:
     - 'charts/everest/**'
 
 jobs:
+  helm-unittest:
```
valmiranogueira: can we call it unit-test to align with the other job stage?

charts/pxc-db/tests/cluster_test.yaml
```diff
@@ -0,0 +1,210 @@
+suite: pxc-db cluster
+templates:
+  - templates/cluster.yaml
+release:
+  name: unit
+  namespace: test-ns
+tests:
+  - it: renders the default cluster
```
valmiranogueira: Instead of check field per field, can we have equal in metadata and spec? With all default generated fields?

charts/pxc-db/tests/cluster_test.yaml
```diff
@@ -0,0 +1,210 @@
...
+          value: unit-pxc-db-s3-backup1
+        documentIndex: 0
+      - notExists:
+          path: spec.backup.storages.backup1.s3.credentialsAccessKey
+        documentIndex: 0
+      - notExists:
+          path: spec.backup.storages.backup1.s3.credentialsSecretKey
+        documentIndex: 0
+
+  - it: renders configured backup and pitr options
```
valmiranogueira: Same as previously, can all backup fields be checked with equal: spec.backup? So we make sure all generated fields are consistent with what is expected

valmiranogueira (changes requested):

valmiranogueira (changes requested):
We will also need .helmignore, to avoid tests on helm charts (tgz)

charts/pxc-db/tests/cluster_test.yaml
pooknull: https://github.com/percona/percona-helm-charts/pull/819/commits/338fdac88a3ebdd84cfcf3378d94e3afacbce7bd

charts/pxc-db/tests/cluster_test.yaml
pooknull: https://github.com/percona/percona-helm-charts/pull/819/commits/338fdac88a3ebdd84cfcf3378d94e3afacbce7bd

.github/workflows/test.yaml
pooknull: https://github.com/percona/percona-helm-charts/pull/819/commits/338fdac88a3ebdd84cfcf3378d94e3afacbce7bd

eleo007 (approved):

PR #777, opened 2026-01-23
[psmdb-db]: pass replset.configuration through tpl function

KyriosGN0:
fixes #751

PR #842, opened 2026-04-17
[psmdb-db]: pass pmm.mongodParams and mongosParams through tpl function

DrFaust92:
## Description

Pass `pmm.mongodParams` and `pmm.mongosParams` through the `tpl` function so they can reference Helm template variables (e.g. `.Values.global.environment`).

This allows users to avoid duplicating the full parameter string per cluster and instead template dynamic values like environment name:

```yaml
# values.yaml
pmm:
  mongodParams: "--environment={{ .Values.global.environment }} --collector.diagnosticdata"
  mongosParams: "--environment={{ .Values.global.environment }}"
```

Follows the same pattern as #777 which did this for `replset.configuration`.

## Changes

- `cluster.yaml`: wrap `pmm.mongodParams` and `pmm.mongosParams` with `tpl` (same as `replset.configuration` in #777)

### DIFF
--- a/charts/psmdb-db/templates/cluster.yaml
+++ b/charts/psmdb-db/templates/cluster.yaml
@@ -128,10 +128,10 @@ spec:
     image: "{{ .Values.pmm.image.repository }}:{{ .Values.pmm.image.tag }}"
     serverHost: {{ .Values.pmm.serverHost }}
     {{- if .Values.pmm.mongodParams }}
-    mongodParams: {{ .Values.pmm.mongodParams }}
+    mongodParams: {{ tpl .Values.pmm.mongodParams $ }}
     {{- end }}
     {{- if .Values.pmm.mongosParams }}
-    mongosParams: {{ .Values.pmm.mongosParams }}
+    mongosParams: {{ tpl .Values.pmm.mongosParams $ }}
     {{- end }}
     {{- if .Values.pmm.resources }}
     resources:


charts/psmdb-db/templates/cluster.yaml
```diff
@@ -128,7 +128,7 @@ spec:
     image: "{{ .Values.pmm.image.repository }}:{{ .Values.pmm.image.tag }}"
     serverHost: {{ .Values.pmm.serverHost }}
     {{- if .Values.pmm.mongodParams }}
-    mongodParams: {{ .Values.pmm.mongodParams }}
+    mongodParams: {{ tpl .Values.pmm.mongodParams $ }}
     {{- end }}
     {{- if .Values.pmm.mongosParams }}
     mongosParams: {{ .Values.pmm.mongosParams }}
```
egegunes: should we do the same for `mongosParams`?

DrFaust92: [egegunes](https://github.com/egegunes) id rather scope ticket to smallest change the gets percona team to merge this, i dont find appying it to both fields

egegunes: i find it wrong to have mongodParams templated but mongosParams not, i think we should do it to merge this

DrFaust92: Done, addressed in latest push.

egegunes (approved):

DrFaust92:
@egegunes thanks for the review! This is approved and up to date with the `release-psmdb-1.23.0` base (no conflicts). Could you merge it when you get a chance? Thanks!

jvpasinatto (approved):

PR #843, opened 2026-04-25
K8SPSMDB-1742 Publish charts as OCI artifacts to GHCR

### LINKED ISSUE #721: Support OCI artifact releases for charts
to allow using a proxy to pull charts from internal registry

example: https://github.com/runatlantis/helm-charts/pull/501

this should be very simple as GH action already used. i can implement this, but want to hear from maintainers

DrFaust92:
https://perconadev.atlassian.net/browse/K8SPSMDB-1742

## Summary
- After `chart-releaser-action` runs, log in to `ghcr.io` and push every newly-released package from `.cr-release-packages/*.tgz` to `oci://ghcr.io/percona/percona-helm-charts`.
- Document the OCI install path in the top-level README.

Closes #721

## Test plan
- [ ] Merge to `main` triggers `Release Charts`; verify chart-releaser still creates the GitHub release as before.
- [ ] Verify the new "Push charts to GHCR" step pushes only the newly-released packages and that the resulting artifacts appear at https://github.com/orgs/percona/packages?repo_name=percona-helm-charts.
- [ ] Confirm `helm pull oci://ghcr.io/percona/percona-helm-charts/<chart> --version <version>` works.

### DIFF
--- a/.github/workflows/everest-release.yaml
+++ b/.github/workflows/everest-release.yaml
@@ -17,10 +17,6 @@ jobs:
       # Is set if the VERSION corresponds to an RC.
       IS_RC: 0
     runs-on: ubuntu-latest
-    permissions:
-      contents: write # to push chart release and create a release (helm/chart-releaser-action)
-      packages: write # needed for ghcr access
-      id-token: write # needed for keyless signing
     steps:
       - name: Set environment variables
         run: |
@@ -69,18 +65,3 @@ jobs:
           skip_existing: true
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-
-      # see https://github.com/helm/chart-releaser/issues/183
-      - name: Login to GitHub Container Registry
-        uses: docker/login-action@v4.1.0
-        with:
-          registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Push charts to GHCR
-        run: |
-          shopt -s nullglob
-          for pkg in .cr-release-packages/*.tgz; do
-            helm push "${pkg}" "oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/percona-helm-charts"
-          done
--- a/.github/workflows/everest-release.yaml
+++ b/.github/workflows/everest-release.yaml
@@ -17,6 +17,10 @@ jobs:
       # Is set if the VERSION corresponds to an RC.
       IS_RC: 0
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # to push chart release and create a release (helm/chart-releaser-action)
+      packages: write # needed for ghcr access
+      id-token: write # needed for keyless signing
     steps:
       - name: Set environment variables
         run: |
@@ -65,3 +69,18 @@ jobs:
           skip_existing: true
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+      # see https://github.com/helm/chart-releaser/issues/183
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v4.1.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push charts to GHCR
+        run: |
+          shopt -s nullglob
+          for pkg in .cr-release-packages/*.tgz; do
+            helm push "${pkg}" "oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/percona-helm-charts"
+          done
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,6 +11,10 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # to push chart release and create a release (helm/chart-releaser-action)
+      packages: write # needed for ghcr access
+      id-token: write # needed for keyless signing
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -51,3 +55,18 @@ jobs:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
         with:
           skip_existing: true
+
+      # see https://github.com/helm/chart-releaser/issues/183
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v4.1.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push charts to GHCR
+        run: |
+          shopt -s nullglob
+          for pkg in .cr-release-packages/*.tgz; do
+            helm push "${pkg}" "oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/percona-helm-charts"
+          done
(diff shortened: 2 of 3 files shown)


.github/workflows/everest-release.yaml
egegunes, line 1: @DrFaust92 I don't want to change everest helm chart in this PR. It's not maintained by us and I don't want to change it without maintainer's approval. also I think everest now uses a different repo: https://github.com/openeverest/helm-charts

egegunes (changes requested):

.github/workflows/everest-release.yaml
DrFaust92: Done, reverted in 5321431. Agreed — let's leave everest alone here.

egegunes (approved):
@jvpasinatto I think we need to test OCI artifact before merging

DrFaust92:
[egegunes](https://github.com/egegunes) any news on this? we would love to have this and this additive and doesnt replace existing chart publishing

jvpasinatto:
Hi @DrFaust92
Could we also add a check to avoid pushing a chart version that already exists in GHCR? We already do this for the chart-releaser step with `skip_existing: true` a similar guard here would prevent accidentally overwriting an existing OCI chart version. The rest looks fine to me, thanks!

DrFaust92:
> Hi @DrFaust92 Could we also add a check to avoid pushing a chart version that already exists in GHCR? We already do this for the chart-releaser step with `skip_existing: true` a similar guard here would prevent accidentally overwriting an existing OCI chart version. The rest looks fine to me, thanks!

Done!

jvpasinatto:
@ademidoff As the release job is also used by PMM charts, is it ok for you?

jvpasinatto (dismissed):

DrFaust92:
@jvpasinatto @egegunes I've rebased this onto the latest `main` to resolve conflicts (the OpenEverest cleanup and the new `Testing Charts` README section) — CI is green again. The rebase dismissed the prior approval, so it now shows `REVIEW_REQUIRED`. Could you re-approve and merge when you have a moment? Thanks!

jvpasinatto (dismissed):

.github/workflows/release.yaml
```diff
@@ -46,3 +50,18 @@ jobs:
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push charts to GHCR
+        run: |
+          shopt -s nullglob
+          for pkg in .cr-release-packages/*.tgz; do
+            helm push "${pkg}" "oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/percona-helm-charts"
```
ademidoff: What if one the `heml push` fails within this loop? That's going to leave the chart registry in inconsistent state.

This PR relies on the happy path only, I think there should be a better way.

ademidoff (changes requested):

.github/workflows/release.yaml
DrFaust92: Updated the push step to address this: each `helm push` now retries transient failures (3 attempts with backoff), and the loop continues pushing the remaining packages instead of aborting on the first error. Any chart that still fails is collected and the job exits non-zero at the end, so partial failures are visible and the run can be retried. Re-runs are idempotent since OCI chart versions are immutable. Done in 1e71d75.

.github/workflows/release.yaml
```diff
@@ -46,3 +50,33 @@ jobs:
...
+
+      - name: Push charts to GHCR
+        run: |
+          shopt -s nullglob
+          repo="oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/percona-helm-charts"
+          failed=()
+          for pkg in .cr-release-packages/*.tgz; do
+            pushed=false
+            for attempt in 1 2 3; do
+              if helm push "${pkg}" "${repo}"; then
```
egegunes, line 70: commit message says "Re-runs are idempotent since OCI chart versions are immutable" but does this command exit with zero in that case?

DrFaust92: Good catch — the commit message wording is wrong. `helm push` to ghcr.io exits 0 here even when the version already exists: ghcr OCI tags are mutable by default, so a re-push just re-points the tag to the same digest and succeeds. So the step is idempotent, but *not* because versions are immutable — if anything an immutable tag would make the re-push fail, exhaust the 3 retries, and `exit 1`. The real reason it's safe is that the registry accepts re-pushing identical content.

(`chart-releaser-action` with `skip_existing: true` still packages every chart into `.cr-release-packages/`, so this step does re-run `helm push` for already-released charts on every run — it just succeeds.)

I'll reword the commit message to drop the "immutable" claim.

DrFaust92:
Gentle ping on this 🙏 — CI is green and the branch is up to date with `main`. The only thing holding merge is the stale `changes requested` review from before the June rebase. This is additive (it doesn't touch the existing chart-releaser publishing) and lets consumers verify charts via OCI/GHCR provenance, which is a real supply-chain-security win for us. @egegunes @jvpasinatto could one of you re-review and merge when you get a chance? Happy to rebase again if anything's drifted. Thanks!

gkech:
ticket : https://perconadev.atlassian.net/browse/K8SPSMDB-1742

jvpasinatto (approved):

egegunes:
@ademidoff could you please re-review?

egegunes (approved):

ademidoff (approved):

raghuvamsi-d:
Quick query, I tried to pull the latest charts that are exported to this repo. using oci chart pull method. It is saying I am not authorised to do so. Kindly check? Or is this expected [repository is private] ?
@DrFaust92 , @ademidoff 
```
helm pull oci://ghcr.io/percona/percona-helm-charts/psmdb-db --version 1.23.0
Error: failed to perform "FetchReference" on source: GET "https://ghcr.io/v2/percona/percona-helm-charts/psmdb-db/manifests/1.23.0": response status code 403: denied: requested access to the resource is denied
```

DrFaust92:
The first post-merge `Release Charts` run (from the PSMDB 1.23.0 release, #808) executed the new **Push charts to GHCR** step successfully and created three packages:

- `ghcr.io/percona/percona-helm-charts/psmdb-db:1.23.0`
- `ghcr.io/percona/percona-helm-charts/psmdb-operator:1.23.0`
- `ghcr.io/percona/percona-helm-charts/psmdb-operator-crds:1.23.0`

They're currently **private**, though — anonymous pulls return `401`:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' https://ghcr.io/v2/percona/percona-helm-charts/psmdb-db/tags/list
401
```

New GHCR packages are created **private by default** and there's no push-time flag to change that, so until visibility is flipped the OCI install path documented in the README won't work for end users:

```console
helm install [RELEASE_NAME] oci://ghcr.io/percona/percona-helm-charts/[CHART_NAME]   # currently: unauthorized
```

**What's needed — one-time, by a `percona` org owner.** For each package at https://github.com/orgs/percona/packages?repo_name=percona-helm-charts → **Package settings**:

1. **Change visibility → Public**.
2. Under **Manage Actions access**, confirm this repo is linked (so future versions keep inheriting write access + visibility).

Visibility is per-package, not per-version, so once a package is public every subsequent version this workflow pushes inherits it — this only has to be done the first time a chart is published. Note that any *new* chart added later will again start private and need the same one-time flip.

jvpasinatto:
Hi @DrFaust92 , we are already aware and looking into it, thanks!

PR #848, opened 2026-05-11
PMM-15029 fix ha password generation

4nte:
[PMM-15029](https://perconadev.atlassian.net/browse/PMM-15029)

Fixes Grafana failing to connect to the internal HA Postgres when the auto-generated user password contained special characters.

- Switched `gfuser` / `pmmuser` password type from `ASCII` to `AlphaNumeric` so generated passwords no longer include URI-reserved chars
- Replaced `GF_DATABASE_URL` with discrete `GF_DATABASE_*` env vars in `statefulset.yaml`, since Grafana parses `GF_DATABASE_URL` as a URI and breaks on any special char that slips through.

Together these remove both the source and the trigger of the bug.

[PMM-15029]: https://perconadev.atlassian.net/browse/PMM-15029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charts/pmm-ha/templates/statefulset.yaml
```diff
@@ -202,8 +202,16 @@ spec:
...
+            - name: GF_DATABASE_TYPE
+              value: "postgres"
+            - name: GF_DATABASE_HOST
+              value: "{{ .Release.Name }}-pg-db-ha.{{ .Release.Namespace }}.svc.cluster.local:5432"
+            - name: GF_DATABASE_NAME
+              value: "grafana"
+            - name: GF_DATABASE_USER
+              value: "$(GF_USERNAME)"
+            - name: GF_DATABASE_PASSWORD
+              value: "$(GF_PASSWORD)"
```
maxkondr:
```suggestion
            - name: GF_DATABASE_USER
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.secret.name }}
                  key: GF_USERNAME
            - name: GF_DATABASE_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.secret.name }}
                  key: GF_PASSWORD
```

charts/pmm-ha/values.yaml
```diff
@@ -35,11 +35,6 @@ pmmEnv:
   # PMM_CLICKHOUSE_PASSWORD: "chpass"
   # PMM_POSTGRES_ADDR: "pmmhak8s-pg-db-ha.pmmhak8s.svc.cluster.local:5432"
   PMM_POSTGRES_SSL_MODE: "require"
-  # GF_DATABASE_TYPE: ""
-  # GF_DATABASE_HOST: ""
-  # GF_DATABASE_NAME: ""
-  # GF_DATABASE_USER: ""
-  # GF_DATABASE_PASSWORD: ""
```
4nte: These are rendered by statefulset template and not configurable directly through chart values / pmmEnv

charts/pmm-ha/templates/statefulset.yaml
```diff
@@ -188,22 +188,26 @@ spec:
...
-            - name: GF_DATABASE_URL
-              value: "postgres://$(GF_USERNAME):$(GF_PASSWORD)@{{ .Release.Name }}-pg-db-ha.{{ .Release.Namespace }}.svc.cluster.local:5432/grafana"
+            - name: GF_DATABASE_TYPE
+              value: "postgres"
+            - name: GF_DATABASE_HOST
+              value: "{{ .Release.Name }}-pg-db-ha.{{ .Release.Namespace }}.svc.cluster.local:5432"
+            - name: GF_DATABASE_NAME
+              value: "grafana"
+            - name: GF_DATABASE_USER
+              value: {{ index .Values "pg-db" "grafanaUserName" | quote }}
```
maxkondr, line 205: why user is not in secret as well?

4nte: It's just that Grafana username is missing from the secret, and it's an established pattern in this file, for example:
```yaml
- name: PMM_POSTGRES_USERNAME
        value: {{ index .Values "pg-db" "pmmUserName" | quote }}
```

4nte: Do you think we should refactor this so that gf username (and pg username) are included in the secret, and referenced from there, instead of chart values?

maxkondr: I think - yes, username is part of credentials. I see no reason we shall show it explicitly

4nte: I think this is arbitrary and comes down to a convention we agree to follow, but it's not that simple either.

Categorically, usernames are not sensitive info, meaning we wouldn't redact them from logs (unlike passwords). 
Furthermore, chart values are the the source of truth for the usernames and would still have to remain in chart values due to how the chart is setup, for example **pg-db** chart references pg user through chart values at render time (secret is not created yet). 
So including username in the secret would just duplicate it, yet it would remain exposed in chart values regardless (if that's the motivation).

I think the biggest issue is if today we switch to reading usernames from secrets, we break existing users that upgrade from older chart versions because their secrets are missing the `user` key.
To support this we need to add a fallback for when `user` key is not present in the secret to read from values; Having to introduce this fallback just to support a switch from one convention to another seems unjustified to me.

maxkondr (dismissed):

charts/pmm-ha/values.yaml
```diff
@@ -605,7 +604,7 @@ pg-db:
         - pmm-managed
       options: "CREATEDB CREATEROLE"
       password:
-        type: ASCII
+        type: AlphaNumeric
```
ademidoff: 👍

charts/pmm-ha/values.yaml
```diff
@@ -614,7 +613,7 @@ pg-db:
         - grafana
       options: "CREATEDB CREATEROLE"
       password:
-        type: ASCII
+        type: AlphaNumeric
```
ademidoff: I don't think we should disallow special characters in the creds, should we?

4nte: Agreed, there were two ways to fix the issue, special chars should work fine now that password is provided via `GF_DATABASE_PASSWORD` which doesn't blow up when special chars are present. I will revert this

charts/pmm-ha/templates/statefulset.yaml
maxkondr: the secrets will be created during applying resources from new chart version. new keys will appear in secret and later used in STS.

ademidoff (approved):

JiriCtvrtka (approved):

maxkondr (approved):

PR #855, opened 2026-05-20
K8SPG-995 update versions for PGO release 3.0.0

valmiranogueira:


charts/pg-operator/README.md
```diff
@@ -9,7 +9,7 @@ A job will be created based on `helm` `install`, `upgrade`, or `uninstall`. Afte
 job has completed the RBAC will be cleaned up.
 
 ## Pre-requisites
-* Kubernetes 1.32+
+* Kubernetes 1.31+
```
eleo007, line 12: Why 1.31 but not 1.33?

valmiranogueira: So we test on Openshift that is 4.18, related Kubernetes version is 31 and not 33. Considering this, this is the oldest K8s version we tested in this release.

jvpasinatto (approved):

PR #862, opened 2026-06-04
fix(pmm): set gRPC service to ClusterIP to avoid duplicate nodePort conflict

### LINKED ISSUE #723: PMM Chart - `monitoring-service-grpc1` duplicates nodePort values causing deployment failure
__Chart Version:__ pmm-1.4.14 (PMM 3.5.0)

__Description:__ When `ingress.enabled: true` and `service.type: NodePort` with explicit `nodePort` values defined, the chart fails to deploy because both `monitoring-service` and `monitoring-service-grpc` services attempt to use the same nodePort values.

__Root Cause:__ In `templates/service.yaml`, the gRPC service (created when ingress is enabled) uses the exact same ports configuration as the main service:

```yaml
{{- if .Values.ingress.enabled }}
---
apiVersion: v1
kind: Service
metadata:
  name: {{ .Values.service.name }}-grpc
spec:
  type: {{ $serviceType | default "ClusterIP" }}
  {{- with .Values.service.ports }}
  ports:
    {{- toYaml . | nindent 8 }}  # <-- This copies nodePort values verbatim
  {{- end }}
```

__Steps to Reproduce:__

1. Deploy PMM chart with:

```yaml
service:
  type: NodePort
  ports:
    - port: 80
      nodePort: 32400
      ...
    - port: 443
      nodePort: 32401
      ...
ingress:
  enabled: true
```

2. Observe error: `Service "monitoring-service-grpc" is invalid: spec.ports[0].nodePort: Invalid value: 32400: provided port is already allocated`

Note: attempted to create Jira ticket but could not get it to allow me to create a ticket for this

somaz94:
The `-grpc` service created when `ingress.enabled=true` inherited the main
`service.type`. With `service.type: NodePort` and explicit
`service.ports[].nodePort` values, both `monitoring-service` and
`monitoring-service-grpc` requested the same nodePorts, so the second service
failed to deploy (a nodePort can only be claimed by one service).

The `-grpc` service is only referenced as an internal Ingress backend (see
`templates/ingress.yaml`), so it never needs to be NodePort/LoadBalancer. This
sets it to `ClusterIP` and drops the `nodePort` field from its ports, which also
avoids a redundant second LoadBalancer when `service.type: LoadBalancer`.

Validation (helm template, chart pmm-1.8.1):
- Default (ingress disabled): only `monitoring-service` rendered, unchanged.
- NodePort + explicit nodePort + ingress enabled: `monitoring-service` keeps its
  nodePorts; `monitoring-service-grpc` renders as ClusterIP with no nodePort — no conflict.
- service.type=ClusterIP + ingress enabled: both ClusterIP, unchanged.
- `helm lint` and `ct lint` pass.

closes: #723

---

Drafted with Claude Code (Opus 4.8) and manually reviewed.

### DIFF
--- a/charts/pmm/Chart.yaml
+++ b/charts/pmm/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: pmm
 description: A Helm chart for Percona Monitoring and Management (PMM)
 type: application
-version: 1.9.1
+version: 1.9.2
 appVersion: "3.9.1"
 home: https://github.com/percona/pmm
 maintainers:
--- a/charts/pmm/templates/service.yaml
+++ b/charts/pmm/templates/service.yaml
@@ -40,10 +40,15 @@ metadata:
     {{- toYaml . | nindent 8 }}
   {{- end }}
 spec:
-  type: {{ $serviceType | default "ClusterIP" }}
+  # The gRPC service is only consumed internally as an Ingress backend, so it is
+  # always ClusterIP. Inheriting the main service type (e.g. NodePort) would make
+  # both services request the same nodePort values and fail to deploy.
+  type: ClusterIP
   {{- with .Values.service.ports }}
   ports:
-    {{- toYaml . | nindent 8 }}
+    {{- range . }}
+    {{- toYaml (list (omit . "nodePort")) | nindent 8 }}
+    {{- end }}
   {{- end }}
   selector:
     {{- include "pmm.selectorLabels" . | nindent 4 }}


somaz94:
Gentle ping 🙏 — green and ready for review. This sets the PMM gRPC service to ClusterIP to avoid a duplicate NodePort allocation. Happy to address any feedback. Thanks!

somaz94:
Friendly ping on this one — it's a small fix for the duplicate nodePort conflict on the PMM gRPC service, and CI is green. Anything I can adjust to help it get a review? Thanks!

somaz94:
Checking in on this one. It only changes the pmm gRPC service to ClusterIP so it stops competing with the main service for a nodePort. It is green. Happy to adjust if you would prefer a different fix.

PR #868, opened 2026-06-12
PMM-15151: make PG and ClickHouse operators watch all namespaces

theTibi:
Running a second `pmm-ha` instance in another namespace (e.g. a DR restore target) failed in two independent ways. This PR fixes both.

## 1. The operators only watched their own namespace

`pmm-ha-dependencies` installed the PostgreSQL (Percona PGO) and ClickHouse (Altinity) operators scoped to their own namespace, so they did not reconcile `PerconaPGCluster` / `ClickHouseInstallation` CRs created elsewhere. Installing `pmm-ha` into a second namespace left PostgreSQL and ClickHouse with no pods, while only VictoriaMetrics came up — its operator is already cluster-scoped.

- **pg-operator**: `watchAllNamespaces=true`. Renders `WATCH_NAMESPACE=""` (all namespaces) and switches the operator RBAC from a namespaced `Role`/`RoleBinding` to `ClusterRole`/`ClusterRoleBinding`.
- **altinity-clickhouse-operator**: RBAC is already cluster-wide, but the operator only watched its own namespace; set `WATCH_NAMESPACES=.*` so it watches every namespace. (An empty watch list, documented upstream as "all", does not take effect in 0.25.4 — see the inline notes in `values.yaml` for why, and why a plain namespace list is *not* a safe way to narrow it.) Mirrored on the metrics-exporter sidecar, which reads its own env.

## 2. A distinct release name produced names that don't resolve

A second install must use a different Helm release name (the `pmm-ha` chart's `ClusterRole`/`ClusterRoleBinding` are named after `pmm.fullname`). But three places derived names from `Release.Name` or `pmm.fullname` inconsistently, and `pmm.fullname` equals `Release.Name` only when the release name already contains the chart name — `pmm-ha2` does, `pmm-2` does not. So all three were invisible under the default release name and broke under any other:

- **`pmm.haPeers`** used `Release.Name`, but the pods are `<fullname>-<ordinal>`. Release `pmm-2` yielded peers `pmm-2-0...` for pods named `pmm-2-pmm-ha-0` — the peers never resolve and the HA memberlist panics on startup. Now uses `pmm.fullname`.
- **HAProxy init script** built the same pod hostnames from `Release.Name`. Same fix.
- **ClickHouse + Keeper scrape jobs** matched the pod `chi`/`chk` label against `pmm.fullname`, but those CRs are named from `Release.Name` — so neither job matched and ClickHouse/Keeper metrics were silently dropped. Fixed by matching on `Release.Name`, which is what the label carries. (Renaming the CRs instead would recreate the ClickHouse cluster on existing releases.)

## Compatibility

No change for existing single-namespace installs using the standard `pmm-ha` release name — `pmm.fullname == Release.Name` there, so every name above renders identically.

Two things do change for everyone, both documented in the chart README:

- Installing `pmm-ha-dependencies` now requires permission to create cluster-scoped RBAC. Single-namespace users who want least privilege can set `pg-operator.watchAllNamespaces: false`.
- Existing operator installs stay namespace-scoped until `helm upgrade` — and the operators must be upgraded **before** installing into a second namespace. Doing it in the other order is not recoverable by retrying: PMM boots with no PostgreSQL, Grafana initializes with the default `admin` password, and `PMM_ADMIN_PASSWORD` is only applied at first initialization.

## Related

Merge together with (and before) #865, which documents the multi-namespace install path this PR makes work. Both target `PMM-HA-GA`.


charts/pmm-ha-dependencies/values.yaml
```diff
@@ -16,9 +16,26 @@ victoria-metrics-operator:
 ## The altinity-clickhouse-operator manages ClickHouse clusters
 altinity-clickhouse-operator:
   enabled: true
+  # The operator's RBAC is already cluster-wide (rbac.namespaceScoped defaults to false),
+  # but it only watches its own namespace unless told otherwise — so ClickHouseInstallations
+  # in other namespaces (e.g. a DR restore target) are never reconciled. Make it watch all
+  # namespaces. NOTE: an empty watch list ("all namespaces" per the docs) did NOT take
+  # effect in 0.25.4 during testing (the operator stayed on its own namespace), so use a
```
ademidoff: The comment reads like an upstream defect, but this is the documented-in-code default, not a 0.25.4 bug. From `applyDefaultWatchNamespace()` in `pkg/apis/clickhouse.altinity.com/v1/type_configuration_chop.go` (Altinity/clickhouse-operator @ 0.25.4):

> 1. Operator runs in 'kube-system' namespace - assume this is global installation, need to watch ALL namespaces
> 2. Operator runs in other (non kube-system) namespace - assume this is local installation, watch this namespace only

So "empty watch list == all namespaces" only holds when the operator is installed into `kube-system`; anywhere else it deliberately pins `Watch.Namespaces.Include` to its own namespace. As written, the note invites a future "let's retest this upstream and drop the workaround" revert. Suggest restating it as the actual rule, e.g.:

```
  # NOTE: leaving the watch list empty does NOT mean "all namespaces" here — the operator
  # only infers "all" when it runs in kube-system; installed anywhere else it pins the watch
  # list to its own namespace (see applyDefaultWatchNamespace() upstream). Hence the regexp.
```

For what it's worth, `.*` is the right lever and it does work as intended: `GetInformerNamespace()` returns `metav1.NamespaceAll` unless `Include` holds exactly one entry matching `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$` — `.*` fails that check, so the informers stay cluster-wide, and `isNamespaceIncluded()` matches via `util.MatchArrayOfRegexps`.

charts/pmm-ha-dependencies/values.yaml
```diff
@@ -16,9 +16,26 @@ victoria-metrics-operator:
...
+  # regexp that matches every namespace. Validate after deploy; alternatively set this to a
+  # comma-separated list/regexp of specific DR target namespaces.  [PMM-15151]
```
ademidoff: This fallback advice is a footgun as phrased, because of the `GetInformerNamespace()` logic above:

- **Exactly one** plain namespace name (e.g. `pmm-dr`) → `Include.Len() == 1` and it matches the label regexp, so the informers get scoped to *that namespace only*. The operator then stops watching its own namespace and the primary PMM install breaks.
- **Two or more** entries (e.g. `pmm,pmm-dr`) → informers fall back to `NamespaceAll` and the regexp filter narrows reconciliation, so this does work — but it still needs the cluster-wide list/watch RBAC.

Either drop the sentence, or spell out the constraint: two or more entries, and the operator's own namespace must be one of them.

ademidoff (commented):
Reviewed the diff against PMM-15151, plus the Altinity operator 0.25.4 and pg-operator 2.8.0 sources, and rendered both charts locally. Skipping the points Copilot already raised (title/description scope, and the blast radius of defaulting both operators to cluster-wide) — I agree with those and won't repeat them.

Three inline comments below, plus one docs item:

### Docs aren't updated

- `charts/pmm-ha-dependencies/README.md` — the "Configuration" section still shows a bare `enabled: true` snippet for both operators. It should show the new `watchAllNamespaces` / `WATCH_NAMESPACES` settings and note the privileges they require (the "Permission issues" section already mentions cluster-admin, so it's a small edit).
- Nothing documents the DR install flow, and there's a hard constraint worth stating explicitly: `charts/pmm-ha/templates/clusterrole.yaml` and `charts/pmm-ha/templates/clusterrolebinding.yaml` name their cluster-scoped objects after `pmm.fullname`, so a second install **must** use a different release name or it collides with the first on those two objects. That's exactly what the `pmm.haPeers` / HAProxy fixes in this PR unblock, so it belongs in the README rather than only in the commit messages.
- Minor: the `[PMM-15151]` tags in a user-facing `values.yaml` don't mean much to chart consumers; the git history already carries that reference.

One caveat: I verified by reading the operator sources and rendering the charts, not by deploying — the cross-namespace reconciliation result still rests on your validation.

charts/pmm-ha-dependencies/values.yaml
```diff
@@ -16,9 +16,26 @@ victoria-metrics-operator:
...
+  operator:
+    env:
+      - name: WATCH_NAMESPACES
+        value: ".*"
```
ademidoff, line 39: Low-priority: the metrics-exporter sidecar doesn't get this variable. The Altinity `Deployment` splices `.Values.metrics.env` into that second container separately, so the rendered output carries `WATCH_NAMESPACES` on the operator container only.

`cmd/metrics_exporter/app/metrics_exporter.go` calls `DiscoveryWatchedCHIs`, which uses the same `GetInformerNamespace()` — so on exporter startup it discovers CHIs in its own namespace only. It self-heals, because the operator pushes watched CHIs to `http://127.0.0.1:8888/chi` (`InformMetricsExporterAboutWatchedCHI` in `pkg/controller/chi/controller.go`), but a CHI in the DR namespace stays unmonitored until the next reconcile touches it.

Mirroring the same env under `metrics.env` closes the gap:

```yaml
  metrics:
    env:
      - name: WATCH_NAMESPACES
        value: ".*"
```

theTibi:
### ⚠️ Merge together with #865 — and merge this one first

Retargeted from `main` to `PMM-HA-GA` so both PRs share a base. #865 ("support installing PMM-HA into multiple namespaces") documents the multi-namespace install path that **this** PR makes actually work:

- The `pmm.haPeers` fix here is what makes a distinct release name viable at all. Without it, `helm install pmm-dr percona/pmm-ha` renders pods `pmm-dr-pmm-ha-{0,1,2}` but `PMM_HA_PEERS` = `pmm-dr-{0,1,2}...`, so the peers never resolve and the HA memberlist panics on startup. It only happens to work when the release name already contains the chart name (`pmm-ha2` yes, `pmm-dr` no).
- The `watchAllNamespaces` / `WATCH_NAMESPACES=".*"` defaults here are what get the PG cluster and CHI reconciled in the second namespace.

If #865 merges first, its documented example is broken on both counts, and it links to this PR's `## Multi-namespace support` heading — a dead link until this lands.

Worth calling out in the release notes for anyone with an **existing** `pmm-ha-dependencies` install: these are values defaults, so a `helm upgrade` of the operators is required before installing into a second namespace. Doing it in the other order is a one-way door — PMM boots without PostgreSQL, Grafana initializes with the default `admin` password, and since `PMM_ADMIN_PASSWORD` is only applied at first initialization it stays that way (the `pmm-token-init` job then loops on 401).

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +64,84 @@ victoria-metrics-operator:
...
+helm install pmm-2 percona/pmm-ha --namespace pmm-2 --create-namespace
+```
+
+### Existing installations: upgrade the operators first
+
+The cluster-wide watch is a **values default**, so an operator release installed before this
+version is still namespace-scoped. Upgrade it before installing into a second namespace:
+
+```bash
+helm upgrade pmm-operators percona/pmm-ha-dependencies --namespace pmm
```
4nte: release name `pmm-operators` here should be renamed to `pmm-ha-operators`, to be consistent with the rest of the docs.

4nte (approved):

JiriCtvrtka (approved):

ademidoff (commented):
Review of the multi-namespace change. Two groups below: the new **Multi-namespace support** section documents a path that does not currently work end-to-end, and there are three small template fixes worth folding into this PR.

The operator-scope flips themselves (`WATCH_NAMESPACES: ".*"`, `watchAllNamespaces: true`) are mechanically correct — verified in the render that both reach the operator *and* the metrics-exporter sidecar, and that the subchart appends rather than replaces the default env.

Not commented inline, but worth a look separately: both operators now hold cluster-wide informers with their subcharts' original resource footprints untouched — pg-operator at `limits.memory: 500Mi`, and the ClickHouse operator at `resources: {}` (BestEffort, first evicted under node pressure). That is a new failure mode on large clusters for every user, including single-namespace ones.

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +64,84 @@ victoria-metrics-operator:
...
+Both the ClickHouse and PostgreSQL operators default to watching **all** namespaces (VictoriaMetrics is already cluster-scoped), so `pmm-ha` custom resources are reconciled no matter which namespace they live in. You install this dependencies chart **once** per cluster; the operators then serve every namespace, and you can run more than one `pmm-ha` instance in separate namespaces of the same cluster.
+
+**A second `pmm-ha` install must use a different Helm release name.** The `pmm-ha` chart creates cluster-scoped objects — its `ClusterRole` and `ClusterRoleBinding` are named after `pmm.fullname` (derived from the release name). A second install with the same release name collides with the first on those two objects, even when the two installs are in different namespaces. Use a distinct release name (which also flows into the correct pod/peer DNS), for example:
+
+```bash
+# First instance
+helm install pmm-ha percona/pmm-ha --namespace pmm
+
+# Second instance in another namespace, different release name
+helm install pmm-2 percona/pmm-ha --namespace pmm-2 --create-namespace
```
ademidoff: This command aborts before applying a single object, so the headline example of the new section can't be followed.

`pmm-ha` defaults to `secret.create: false`, and `charts/pmm-ha/templates/pg-user-credentials-secrets.yaml:3` hard-`fail`s at render time when `pmm-secret` is absent — which it always is in a namespace that `--create-namespace` just made. Reproduced verbatim:

```
$ helm template pmm-2 charts/pmm-ha --namespace pmm-2 -f charts/pmm-ha/ci/ci-values.yaml
Error: execution error at (pmm-ha/templates/pg-user-credentials-secrets.yaml:3:4): Secret 'pmm-secret' not found in namespace 'pmm-2'.
```

The repo's own CI hits this and works around it by pre-creating the secret before install (`.github/workflows/pmm-ha-pr-checks.yaml:96-113`). The docs need the equivalent step here — either that pre-create, or whichever of `secret.create=true` / an explicit `--set` you intend as the supported path for a second instance.

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +64,84 @@ victoria-metrics-operator:
...
+>
+> The ClickHouse operator does **not** support scoping to a single extra namespace — a
+> one-entry watch list scopes it to that namespace *only* and breaks the primary install.
+> See the inline notes in `values.yaml`.
+
+## Multi-namespace support
+
+Both the ClickHouse and PostgreSQL operators default to watching **all** namespaces (VictoriaMetrics is already cluster-scoped), so `pmm-ha` custom resources are reconciled no matter which namespace they live in. You install this dependencies chart **once** per cluster; the operators then serve every namespace, and you can run more than one `pmm-ha` instance in separate namespaces of the same cluster.
+
+**A second `pmm-ha` install must use a different Helm release name.** The `pmm-ha` chart creates cluster-scoped objects — its `ClusterRole` and `ClusterRoleBinding` are named after `pmm.fullname` (derived from the release name). A second install with the same release name collides with the first on those two objects, even when the two installs are in different namespaces. Use a distinct release name (which also flows into the correct pod/peer DNS), for example:
```
ademidoff: "A second `pmm-ha` install must use a different Helm release name" is necessary but not sufficient, and the stated collision surface is short on both ends.

**Namespaced objects that collide regardless of release name.** Rendering two releases into the same namespace and intersecting the object names gives six: `pmm-ha-haproxy` and `pmm-ha-haproxy-init-script` (from `haproxy.fullnameOverride: pmm-ha-haproxy`, `charts/pmm-ha/values.yaml:485`), `monitoring-service` (`service.name` default), `pmm-service-account`, `haproxy-tls-secret`, `postgresql-init-extensions`. So a reader who takes the sentence at face value and runs `helm install pmm-2 ... --namespace pmm` either gets an existing-resource abort, or on an adopt/`--force` path silently repoints the first release's HAProxy at the second release's pods. The real rule is **one `pmm-ha` per namespace**, and the text should say that.

**Cluster-scoped objects beyond the two named.** The render also produces `<release>-kube-state-metrics` ClusterRole and ClusterRoleBinding. Worth noting each instance's KSM is unscoped, so every tenant's PMM ingests and displays every other namespace's object state.

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +64,84 @@ victoria-metrics-operator:
...
+>   watchAllNamespaces: false
+> ```
+>
+> The ClickHouse operator does **not** support scoping to a single extra namespace — a
+> one-entry watch list scopes it to that namespace *only* and breaks the primary install.
+> See the inline notes in `values.yaml`.
+
+## Multi-namespace support
+
+Both the ClickHouse and PostgreSQL operators default to watching **all** namespaces (VictoriaMetrics is already cluster-scoped), so `pmm-ha` custom resources are reconciled no matter which namespace they live in. You install this dependencies chart **once** per cluster; the operators then serve every namespace, and you can run more than one `pmm-ha` instance in separate namespaces of the same cluster.
```
ademidoff: The "run more than one `pmm-ha` instance in separate namespaces" claim doesn't hold for node-exporter, and a distinct release name doesn't help.

`charts/pmm-ha/values.yaml:710-711` has `prometheus-node-exporter.enabled: true`, and the subchart runs `hostNetwork: true` on port `9100`. The second instance's DaemonSet pods CrashLoopBackOff on `listen tcp [ip]:9100: bind: address already in use` on every node — permanently, and silently as far as the install is concerned. Instance 2 gets no node/OS metrics.

The chart already treats exactly this collision as fatal elsewhere (`charts/pmm-ha/templates/_helpers.tpl:195` fails the render with "...would collide with OpenShift's node-exporter on host port 9100."), so the second-install case arguably deserves the same guard — or at minimum a note here telling the second instance to set `prometheus-node-exporter.enabled: false`.

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +64,84 @@ victoria-metrics-operator:
...
+```
+
+### Existing installations: upgrade the operators first
+
+The cluster-wide watch is a **values default**, so an operator release installed before this
+version is still namespace-scoped. Upgrade it before installing into a second namespace:
+
+```bash
+helm upgrade pmm-ha-operators percona/pmm-ha-dependencies --namespace pmm
+kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=pg-operator -n pmm --timeout=180s
```
ademidoff: These two `kubectl wait` calls are presented as the gate that proves it's safe to proceed, but they return a false green.

`helm upgrade` on the line above runs without `--wait`, so it returns as soon as the Deployment spec is patched. The only change is env vars, so this is a rolling update with an overlap window — `kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=pg-operator` then matches the **pre-upgrade** pod, which is already Ready, and succeeds in under a second. The user proceeds to install into the second namespace while the still-namespace-scoped operators are running, landing in the state the box at line 138 calls unrecoverable.

`kubectl rollout status deployment/<name> -n pmm --timeout=...` is the primitive that actually waits for the new generation. Two smaller things while you're here: this recipe uses `--timeout=180s` where the existing one elsewhere in the file uses 300s, and it drops victoria-metrics-operator, which the other one waits on.

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +64,84 @@ victoria-metrics-operator:
...
+kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=altinity-clickhouse-operator -n pmm --timeout=180s
+```
+
+Do **not** reinstall this chart into the new namespace — the VictoriaMetrics operator CRDs are
+owned by the first release, and a second install fails with
+`invalid ownership metadata ... release-namespace must equal "<new-namespace>"`.
+
+> **⚠️ Order matters — this one is not recoverable by retrying.** If you install `pmm-ha` into
```
ademidoff, line 195: The symptom described here can't actually occur, so this will send whoever hits the problem off in the wrong direction.

When the ClickHouseInstallation is never reconciled, the unconditional `wait-for-clickhouse` init container (`charts/pmm-ha/templates/statefulset.yaml:65`) loops `while :;` with `sleep 5` and no timeout. The pmm container never starts, so Grafana never initializes and there is no default-`admin`-password state and no `pmm-token-init` 401 loop.

The observable symptom is a PMM pod stuck at `Init:1/2`. Someone reading this box goes hunting for a wrong admin password instead of an unreconciled CHI/PostgresCluster. Suggest rewriting around the init-container stall — the "order matters, and retrying won't fix it" point stands, it's just the evidence that's wrong.

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +64,84 @@ victoria-metrics-operator:
...
+> ```yaml
+> pg-operator:
+>   watchAllNamespaces: false
+> ```
+>
+> The ClickHouse operator does **not** support scoping to a single extra namespace — a
+> one-entry watch list scopes it to that namespace *only* and breaks the primary install.
+> See the inline notes in `values.yaml`.
+
+## Multi-namespace support
```
ademidoff, line 126: This section makes the chart install-once-per-cluster, which invalidates the **Uninstallation Order** section above it (`charts/pmm-ha-dependencies/README.md:46`) and the NOTES.txt — both still written for a single-tenant install and left untouched by this PR.

`helm uninstall pmm-ha-operators --namespace pmm` now tears the pg/ClickHouse/VM operators out cluster-wide, so every *other* namespace's `PerconaPGCluster`, `ClickHouseInstallation` and `VMCluster` silently stop being reconciled. The obvious recovery — reinstalling the deps chart into a surviving namespace — is then blocked: the rendered output carries 22 Helm-templated CRDs, each with `helm.sh/resource-policy: keep` plus `meta.helm.sh/release-namespace: "pmm"`, producing the exact `invalid ownership metadata ... release-namespace must equal "<new-namespace>"` error this PR quotes three lines further down. The only way out is hand-patching annotations on 22 cluster-scoped CRDs.

At minimum the uninstall section needs a "this affects every namespace in the cluster" warning. `charts/pmm-ha/README.md` and both NOTES.txt files also still describe the single-namespace model (and `charts/pmm-ha/README.md:58` still says `pmm-operators` where this PR standardizes on `pmm-ha-operators`).

charts/pmm-ha/templates/vmagent.yaml
```diff
@@ -64,9 +64,12 @@ spec:
...
               - {{ .Release.Namespace }}
       relabel_configs:
-        # Keep only ClickHouse pods from this release
+        # Keep only ClickHouse pods from this release. The label carries the
+        # ClickHouseInstallation name, which clickhouse-cluster.yaml sets from Release.Name
+        # (NOT pmm.fullname) - they differ whenever the release name does not already
+        # contain the chart name, e.g. release "pmm-2" -> CHI "pmm-2", fullname "pmm-2-pmm-ha".
         - source_labels: [__meta_kubernetes_pod_label_clickhouse_altinity_com_chi]
-          regex: '{{ include "pmm.fullname" . }}'
+          regex: '{{ .Release.Name }}'
```
ademidoff: The switch back to `.Release.Name` is right — `charts/pmm-ha/templates/clickhouse-cluster.yaml:19` and `charts/pmm-ha/templates/clickhouse-keeper.yaml:4` do name the CRs from `Release.Name`, and I confirmed CHI `pmm-2` matches `regex: 'pmm-2'` in the render. But the value lands unescaped in a fully-anchored RE2 regex, and Helm permits `.` in release names.

`helm install "pmm.ha" <chart> --dry-run` exits 0, rendering `regex: 'pmm.ha'` where `.` matches any character. In a namespace that also holds a CHI named `pmm-ha` (or `pmmXha`) the job keeps the wrong cluster's pods and QAN metrics get cross-attributed.

```suggestion
          regex: '{{ .Release.Name | regexQuoteMeta }}'
```

Same fix needed at line 105 for `'{{ .Release.Name }}-keeper'`.

charts/pmm-ha/templates/vmagent.yaml
```diff
@@ -96,9 +99,10 @@ spec:
             names:
               - {{ .Release.Namespace }}
       relabel_configs:
-        # Keep only ClickHouse Keeper pods from this release
+        # Keep only ClickHouse Keeper pods from this release. As above, the label carries the
+        # ClickHouseKeeperInstallation name, which clickhouse-keeper.yaml sets from Release.Name.
         - source_labels: [__meta_kubernetes_pod_label_clickhouse_keeper_altinity_com_chk]
-          regex: '{{ include "pmm.fullname" . }}-keeper'
+          regex: '{{ .Release.Name }}-keeper'
```
ademidoff: Same unescaped-`Release.Name`-in-an-anchored-regex issue as line 72 — see that comment for the reproduction.

```suggestion
          regex: '{{ .Release.Name | regexQuoteMeta }}-keeper'
```

charts/pmm-ha/templates/vmagent.yaml
```diff
@@ -64,9 +64,12 @@ spec:
             names:
               - {{ .Release.Namespace }}
       relabel_configs:
-        # Keep only ClickHouse pods from this release
+        # Keep only ClickHouse pods from this release. The label carries the
+        # ClickHouseInstallation name, which clickhouse-cluster.yaml sets from Release.Name
+        # (NOT pmm.fullname) - they differ whenever the release name does not already
+        # contain the chart name, e.g. release "pmm-2" -> CHI "pmm-2", fullname "pmm-2-pmm-ha".
```
ademidoff: These four lines sit inside the `inlineScrapeConfig: |` block scalar, so they aren't chart comments — they become part of the VMAgent CR's `spec.inlineScrapeConfig` string and of the scrape-config Secret the operator generates from it.

Consequences: for release `pmm-ha` the vmagent behaviour is unchanged by this PR, yet `helm upgrade` now writes a changed spec, so the VM operator regenerates the config Secret and reloads all vmagent replicas — for prose that only needed to be in git. It also makes `kubectl diff` / GitOps drift review on the CR noisy, and any future comment containing `{{` or a stray tab becomes a runtime scrape-config parse error instead of a render error.

The equivalent note in `charts/pmm-ha/templates/_helpers.tpl:104-109` is correctly written as `{{- /* ... */}}` and stripped at render — these should match. Same applies to lines 102-103 in the keeper job.

charts/pmm-ha/templates/haproxy-init-configmap.yaml
```diff
@@ -47,7 +47,7 @@ data:
       ready_count=0
       
       for i in $(seq 0 $(({{ .Values.replicas }} - 1))); do
-        pmm_host="{{ .Release.Name }}-$i.{{ .Values.service.name | default "monitoring-service" }}.{{ .Release.Namespace }}.svc.cluster.local"
+        pmm_host="{{ include "pmm.fullname" . }}-$i.{{ .Values.service.name | default "monitoring-service" }}.{{ .Release.Namespace }}.svc.cluster.local"
```
ademidoff, line 50: This fix is correct but won't reach running clusters on upgrade: `haproxy.podAnnotations.pmm.percona.com/config-version` (`charts/pmm-ha/values.yaml:515`) wasn't incremented, and that annotation is the chart's own documented mechanism for this — the comment three lines above it reads "Increment config-version only for other ConfigMap changes that require restart."

The haproxy subchart's `checksum/environment` doesn't cover it either: that checksum is over the subchart's own `configmap.yaml` (which renders from the empty `haproxy.config`), not the chart-owned `pmm-ha-haproxy-init-script`. I diffed the rendered HAProxy Deployment old-vs-new for release `pmm-2` — byte-identical, while the init script changed from `pmm-2-$i...` to `pmm-2-pmm-ha-$i...`.

So `helm upgrade` creates no new ReplicaSet, the running pods keep the pre-upgrade script, and the fix only lands on some unrelated future restart — with `helm upgrade` reporting success throughout. Bumping `config-version` in `charts/pmm-ha/values.yaml:515` is the one-line fix.

Separately: this block re-implements the FQDN construction that `pmm.haPeers` in `charts/pmm-ha/templates/_helpers.tpl` already does — which is why this PR had to fix the same bug in two places. Worth collapsing at some point.

charts/pmm-ha/templates/vmagent.yaml
theTibi: Fixed together with line 72 — see that thread.

charts/pmm-ha/templates/vmagent.yaml
theTibi: Fixed. Applied `regexQuoteMeta` here and at line 105, plus the five other interpolated regexes in this file (136/158/180/202/224) — those go through `pmm.fullname`, so they have the same exposure.

Verified with a server-side render: release `pmm-2` is byte-identical to before; release `pmm.ha` now renders `regex: 'pmm\.ha'` where it previously rendered `'pmm.ha'`.

charts/pmm-ha/templates/vmagent.yaml
theTibi: Right — converted both blocks to `{{- /* … */}}` to match `_helpers.tpl:104`. The rendered diff for release `pmm-2` is now just the six removed comment lines plus the `config-version` bump; the CR spec is otherwise unchanged.

charts/pmm-ha/templates/haproxy-init-configmap.yaml
theTibi: Confirmed and fixed — `pmm.percona.com/config-version` bumped to `"4"` in `charts/pmm-ha/values.yaml:515`.

Agreed on the duplicated FQDN construction; leaving that for a follow-up rather than growing this PR.

charts/pmm-ha-dependencies/README.md
theTibi: Correct. Rewritten to state the rule as **one `pmm-ha` per namespace**, listing all six colliding namespaced objects, plus a note that each instance's kube-state-metrics reads cluster-wide so namespaces aren't a tenancy boundary.

Reproduced the abort — a second release in an occupied namespace fails on the `pmm-ha-haproxy` ServiceAccount.

charts/pmm-ha-dependencies/README.md
theTibi: Correct. The example is removed and the section now states the 9100 constraint, pointing at the `pmm-ha` chart README where #865 has the worked example with `--set prometheus-node-exporter.enabled=false`.

On the render-time guard: I don't think it can mirror `_helpers.tpl:195`. That one is a pure values check. Detecting a node-exporter owned by another release needs `lookup`, which returns empty under `helm template`, `--dry-run=client` and GitOps diff — so the guard would be missing exactly where people would rely on it. The documented flag seems better than a guard that silently doesn't fire.

charts/pmm-ha-dependencies/README.md
theTibi: Reproduced — thanks. The example is removed; per-namespace prerequisites including pre-creating `pmm-secret` live in the `pmm-ha` chart README rather than being duplicated here.

charts/pmm-ha-dependencies/README.md
theTibi: Correct. Switched to `kubectl rollout status deployment -l … --timeout=300s` for all three operators including victoria-metrics-operator, with a comment on why the readiness wait returns a false green.

charts/pmm-ha-dependencies/README.md
theTibi: Confirmed and rewritten. `wait-for-pg-secret` waits on `pg-encryption-key`, which the chart's own `pmm-secret-init-job` creates, so it passes; the pod then parks at `Init:1/2` on `wait-for-clickhouse`, which loops with no timeout. Grafana never starts, so there's no default-password state.

The box now describes the Init stall and gives the two commands to check.

charts/pmm-ha-dependencies/README.md
theTibi: Agreed. Added a warning at the top of **Uninstallation Order** covering both the cluster-wide reconciliation impact and the retained-CRD ownership trap, mirrored in the dependencies `NOTES.txt`.

Also renamed the three remaining `pmm-operators` references (`charts/pmm-ha/README.md:58,160`, `values.yaml:729`).

Confirmed the CRD error live: reinstalling the deps chart under a different release/namespace fails with `invalid ownership metadata … release-namespace must equal`.

theTibi:
All ten inline comments addressed — details in each thread. Two things deliberately not done here:

- **Operator resource limits** — real, but changing subchart resourcing affects every user including single-namespace ones and needs load data this PR doesn't have. Tracking separately.
- **A render-time node-exporter guard** — see the `README.md:111` thread for why `lookup` makes it unreliable.

Verified with `helm template --dry-run=server`: the rendered diff for release `pmm-2` is only the removed comments plus the `config-version` bump; `helm lint` clean on both charts; the node-exporter CI assertions still pass.

charts/pmm-ha-dependencies/README.md
```diff
@@ -31,6 +31,22 @@ helm install pmm-ha percona/pmm-ha --namespace pmm
 
 ## Uninstallation Order
 
+> **⚠️ This chart is installed once per cluster and its operators serve every namespace.**
+> Uninstalling it stops reconciliation of `PerconaPGCluster`, `ClickHouseInstallation` and
+> `VMCluster` resources in **all** namespaces, not just this one — every other `pmm-ha`
+> instance on the cluster is affected. Check for other instances first:
+>
+> ```bash
+> helm list -A -f pmm-ha
```
ademidoff: This won't find the instances it's meant to find. `helm list -f` is a regex over the **release name**, not the chart — so `-f pmm-ha` matches a release named `pmm-ha` and misses one named `pmm-2`, which is exactly the naming convention this document establishes ~100 lines further down ("a distinct release name per instance").

That's the awkward failure mode: it returns a plausible-looking list containing the conventionally-named instance, so the reader concludes there are no others and proceeds with the uninstall. Filtering on the chart column instead:

```suggestion
> helm list -A -o json | jq -r '.[] | select(.chart | test("^pmm-ha-[0-9]")) | "\(.namespace)/\(.name)"'
```

(`test("^pmm-ha-[0-9]")` rather than a `startswith` so it doesn't also match `pmm-ha-dependencies-1.0.0`.)

ademidoff (commented):
Thanks — went through all ten and verified each against a render rather than the replies. Eight are correct as described, and I confirmed the parts that were easy to get wrong:

- `regexQuoteMeta` on all seven: release `pmm.ha` now renders `regex: 'pmm\.ha'`, `pmm-2` byte-identical to before.
- The `{{- /* */}}` conversion doesn't break the block scalar — rendered VMAgent diff pre/post is exactly the six removed comment lines, and `inlineScrapeConfig` still parses (13 jobs).
- `config-version: "4"` does change the haproxy pod-template annotation, so the Deployment actually rolls.
- All six fixed-name objects exist in the render, and `pmm-ha-haproxy` is indeed a ServiceAccount.
- `kubectl rollout status -l …` — I'd expected this to reject a selector, but `rollout status` does accept `--selector`. Command is fine.

Both declines are good calls. The `lookup` reasoning on the node-exporter guard is right — it returns empty under `helm template`, `--dry-run=client` and GitOps diff, so the guard would be absent exactly where people would check. Agreed the documented flag is better. And the FQDN de-duplication is fine as a follow-up.

Two small things the new text introduced, below.

charts/pmm-ha-dependencies/templates/NOTES.txt
```diff
@@ -33,5 +37,7 @@ Once all operators are ready, install the pmm-ha chart:
 
 📚 Documentation: https://docs.percona.com/percona-monitoring-and-management/
 
-⚠️  UNINSTALL ORDER: When uninstalling, remove pmm-ha FIRST, then pmm-ha-dependencies
+⚠️  UNINSTALL ORDER: When uninstalling, remove pmm-ha FIRST, then pmm-ha-dependencies.
+    Because the operators are cluster-wide, uninstalling this chart stops reconciliation for
+    EVERY namespace on the cluster. Check for other instances first: helm list -A -f pmm-ha
```
ademidoff: Same `helm list -f` issue as `charts/pmm-ha-dependencies/README.md:40` — it filters on release name, so it misses any instance not named `pmm-ha`.

For NOTES output the jq pipeline is probably more than you want on one line; pointing at the chart column is enough here:

```suggestion
    EVERY namespace on the cluster. Check the CHART column of `helm list -A` for other
    pmm-ha instances first - they may not be named "pmm-ha".
```

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +80,117 @@ victoria-metrics-operator:
...
+2. **A distinct release name per instance.** The `pmm-ha` chart's `ClusterRole` and
+   `ClusterRoleBinding` are named after `pmm.fullname`, which derives from the release name.
+   Those are cluster-scoped, so two instances sharing a release name collide even when they are
+   in different namespaces. A distinct release name also flows into the correct pod/peer DNS.
+
+The per-instance settings — the cluster-wide monitoring sub-charts (only one instance per
+cluster can run `prometheus-node-exporter`, which binds host port 9100) and the per-namespace
+prerequisites (each namespace needs its own `pmm-secret` **before** you install) — are covered
+in the [`pmm-ha` chart README](../pmm-ha/README.md), which is also where the worked
+two-namespace example lives.
```
ademidoff: The `pmm-secret` half of this pointer is solid — `charts/pmm-ha/README.md` has the full "Creating PMM Secret Manually" section. The other two thirds don't resolve on this branch:

- **No `9100` / second-instance node-exporter content.** `charts/pmm-ha/README.md` covers `prometheus-node-exporter` only for the OpenShift case (`nodeExporter.mode: openshift`, around line 558). There is no mention of host port 9100 or of disabling the DaemonSet for a second instance in the same cluster.
- **No worked two-namespace example.** Nothing in `charts/pmm-ha/README.md` matches `pmm-2`, "second instance", or "multi-namespace".

Both live in #865, which is still open against this same `PMM-HA-GA` base. If this PR merges first, the sentence sends readers to content that isn't there.

Not a lost warning — the 9100 constraint is stated inline right here, which is the part that matters. It's the cross-reference that's premature. Simplest fix is to drop "which is also where the worked two-namespace example lives" and re-add it in #865, so the two PRs can merge in either order.

ademidoff (commented):
One more on the multi-namespace section — RBAC rather than kube-state-metrics. Not a blocker and not something this PR introduced, but it's the same "namespaces aren't a tenancy boundary" point and the note above is the natural place for it. Details inline.

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +80,117 @@ victoria-metrics-operator:
...
+
+> **Note:** each instance runs its own `kube-state-metrics` with cluster-wide read access, so
+> every instance ingests and displays object state for the whole cluster, not just for its own
+> namespace. Separate namespaces are **not** a tenancy boundary here.
```
ademidoff: This note is right, and the same caveat applies to RBAC — more sharply. Worth folding in, because the two statements are currently inconsistent with each other.

`charts/pmm-ha/templates/clusterrole.yaml` grants, with no namespace restriction:

```yaml
resources: [namespaces, endpoints, services, nodes, nodes/proxy, nodes/metrics,
            pods, persistentvolumeclaims, secrets, serviceaccounts]
verbs: [get, watch, list, delete]
```

and `charts/pmm-ha/templates/clusterrolebinding.yaml` binds it cluster-wide to that release's ServiceAccount. So with two instances installed per this section, `pmm-service-account` in namespace `pmm-2` can read **and delete** `pmm-secret` (admin password, `PG_PASSWORD`, `GF_PASSWORD`, ClickHouse and VMAuth credentials), `pg-encryption-key` and `haproxy-tls-secret` in namespace `pmm`, and vice versa. Deleting `pg-encryption-key` also wedges the other instance — `charts/pmm-ha/templates/statefulset.yaml:53` blocks in `wait-for-pg-secret` until it reappears.

To be clear about scope: **this ClusterRole predates the PR and I'm not suggesting you narrow it here.** What changes is that the grant used to be an unused footprint on a single-instance install, and this section now actively recommends the configuration that makes it reachable. So the doc and the RBAC have drifted apart.

Two ways to square it, your call:

1. **Extend this note**, e.g. "…and each instance's ServiceAccount holds cluster-wide read/delete on Secrets, so instances are not isolated from each other's credentials. Do not use separate namespaces as a security boundary between untrusted teams." One sentence, keeps the PR self-consistent.
2. **Separate ticket** to narrow the ClusterRole — `delete` on cluster-wide `secrets` and `serviceaccounts` looks broader than service discovery needs, though I haven't traced which controller wants `delete`, so that needs its own investigation.

I'd lean towards (1) now and (2) later, so this PR doesn't grow.

charts/pmm-ha/values.yaml
```diff
@@ -512,7 +512,7 @@ haproxy:
   # so scaling up/down replicas does NOT require HAProxy restart.
   # Increment config-version only for other ConfigMap changes that require restart.
   podAnnotations:
-    pmm.percona.com/config-version: "3"
+    pmm.percona.com/config-version: "4"
```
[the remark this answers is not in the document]
theTibi: CI does not reject this. `lint-test` passes on this head, and the run log prints
`Version increment checking disabled.` immediately after `Linting charts...` — ct 3.10.1 turns
that check off when charts are named explicitly with `--charts=...`, which this workflow does, so
`--check-version-increment=true` is inert. `.github/ct.yaml` also has `target-branch: main`, not
`PMM-HA-GA`, so the comparison base would be wrong even if it ran.

Leaving `pmm-ha` at 1.6.0 deliberately: 1.6.1 is already published from `main`, and this branch
bumps at release time (`Release PMM 3.9.0`-style commits — the three earlier commits on
`PMM-HA-GA` that touch `charts/pmm-ha` don't touch `Chart.yaml`). A per-PR bump here would just
collide with the other open PRs on the branch.

charts/pmm-ha/templates/haproxy-init-configmap.yaml
```diff
@@ -47,7 +47,7 @@ data:
...
```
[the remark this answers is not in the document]
theTibi: Already done — the description has a dedicated section, "2. A distinct release name produced names
that don't resolve", covering both the `pmm.haPeers` change and this HAProxy init-script change.
Resolving.

charts/pmm-ha-dependencies/values.yaml
```diff
@@ -16,9 +16,43 @@ victoria-metrics-operator:
...
 ## @section PostgreSQL Operator Configuration
 ## The pg-operator manages PostgreSQL clusters
 pg-operator:
   enabled: true
+  # Watch ALL namespaces so PostgreSQL clusters created in other namespaces of the
+  # cluster are reconciled. This also switches the operator
+  # RBAC from a namespaced Role/RoleBinding to a ClusterRole/ClusterRoleBinding. Matches
+  # the VictoriaMetrics operator, which is already cluster-scoped. (Subchart default is
+  # false = the operator watches only its own namespace.)
+  watchAllNamespaces: true
```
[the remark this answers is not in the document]
theTibi: The premise is right — the kind job installs both charts into a single namespace, so the
cluster-wide reconciliation path isn't exercised. Not taking it in this PR, though: a genuine
second-namespace case means waiting on a full PostgreSQL + ClickHouse reconcile on kind, on top of
a job that already runs ~8.5 minutes.

The concrete regression risk in this PR is the release-name-sensitive rendering, and I'm covering
that instead with a render check (`helm template pmm-2 charts/pmm-ha --dry-run=server`, which needs
the cluster only because `pg-user-credentials-secrets.yaml` does an unconditional `lookup` + `fail`
on `pmm-secret`) — see the thread on `_helpers.tpl`.

If we add the integration case later, the cheap version is to create just a minimal
`ClickHouseInstallation` + `PerconaPGCluster` in a second namespace and assert the operators touch
them (status populated / StatefulSet created), rather than waiting for full readiness.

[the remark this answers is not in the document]
theTibi: The CI reasoning doesn't hold — `lint-test` passes, and the log prints
`Version increment checking disabled.` because ct 3.10.1 skips that check when charts are passed
explicitly via `--charts=...` (see the thread on `charts/pmm-ha/values.yaml`).

But the bump is warranted for a different and stronger reason: `.github/workflows/release.yaml`
runs `chart-releaser-action` with `skip_existing: true`, and `pmm-ha-dependencies-1.0.0` is already
published. If this chart reaches `main` still at 1.0.0, the changed RBAC/defaults are **silently
never published** — no error, and users keep the namespace-scoped operators. So I'm bumping
`charts/pmm-ha-dependencies` to 1.1.0 (minor: changed defaults and rendered RBAC).

`charts/pmm-ha` stays at 1.6.0 for the reason in the other thread.

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +81,119 @@ victoria-metrics-operator:
...
+> so instances are not isolated from each other's credentials. Separate namespaces are **not** a
+> tenancy boundary here — do not use them as a security boundary between untrusted teams.
+
+### Existing installations: upgrade the operators first
+
+The cluster-wide watch is a **values default**, so an operator release installed before this
+version is still namespace-scoped. Upgrade it before installing into a second namespace:
+
+```bash
+helm upgrade pmm-ha-operators percona/pmm-ha-dependencies --namespace pmm
```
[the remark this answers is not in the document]
theTibi: Correct, and it fails for exactly the audience this section targets: `charts/pmm-ha/README.md:58`
on `PMM-HA-GA` documents `helm install pmm-operators percona/pmm-ha-dependencies`, so an install
predating this change is named `pmm-operators` and `helm upgrade pmm-ha-operators` aborts with
`Error: release: not found` — loudly rather than silently, but the operators do stay
namespace-scoped.

Fixed in ac33e22: the recipe now starts with `helm list --namespace pmm` and uses
`<your-release-name>`, with a sentence noting earlier docs used `pmm-operators`.

charts/pmm-ha-dependencies/README.md
```diff
@@ -64,15 +81,119 @@ victoria-metrics-operator:
...
+
+# `helm upgrade` runs without --wait, so it returns as soon as the Deployments are patched.
+# This change only touches env vars, so it rolls out with an overlap window: a
+# `kubectl wait --for=condition=ready pod` would match the still-Ready *pre-upgrade* pod and
+# succeed immediately. Wait for the new generation instead.
```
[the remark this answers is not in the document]
theTibi: Correct, and understated — pg-operator is `Recreate` too, not just Altinity:

- `pg-operator-2.8.0/templates/deployment.yaml:18` → `strategy: type: Recreate`
- `altinity-clickhouse-operator-0.25.4/values.yaml:9` → `strategy: type: Recreate`
- victoria-metrics-operator sets no strategy, so it is the only one with an overlap window.

And on the RBAC half: `pg-operator/templates/role.yaml:1,9` and `role-binding.yaml:7,15` are gated
on `watchAllNamespaces`, so the upgrade does swap namespaced Role/RoleBinding for cluster-scoped
ones.

The conclusion survives — with `Recreate` the pre-upgrade pod is still present and still `Ready`
(merely Terminating) when helm returns, so the readiness wait still returns a false green. Comment
rewritten around that in ac33e22; the `rollout status` commands are unchanged (`-l` is supported
there, I checked).

charts/pmm-ha-dependencies/templates/NOTES.txt
```diff
@@ -11,6 +11,10 @@ The following operators have been deployed:
   ✓ PostgreSQL Operator (Percona Distribution)
 {{- end }}
 
+🌐 SCOPE: These operators watch ALL namespaces, so this chart is installed once per cluster.
+   Any pmm-ha instance, in any namespace, is reconciled by them - you do not install this
+   chart again for a second namespace.
```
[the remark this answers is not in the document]
theTibi: Right — it contradicts the opt-out this same PR documents at `README.md:117-119`. Fixed in ac33e22:
the SCOPE block now derives the message from the values, warning per operator when
`pg-operator.watchAllNamespaces` is false or `WATCH_NAMESPACES` is not `.*`, and only claiming
cluster-wide otherwise.

Rendered all four combinations (defaults, pg narrowed, ClickHouse narrowed, both) plus a null-value
check on the subchart keys.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -99,8 +99,15 @@ Generate PMM HA peer list dynamically based on replicas count
...
+{{- $fullname := include "pmm.fullname" . }}
 {{- range $i := until $replicas }}
-  {{- $peer := printf "%s-%d.%s.%s.svc.cluster.local" $.Release.Name $i $serviceName $.Release.Namespace }}
+  {{- /* Peers must use the StatefulSet name (pmm.fullname), not Release.Name: the pods are
+         <fullname>-<ordinal>. pmm.fullname equals Release.Name only when the release name
+         already contains the chart name (e.g. "pmm-ha" or "pmm-ha-2"); otherwise it is
+         "<release>-pmm-ha" (e.g. release "pmm-2" -> pods "pmm-2-pmm-ha-0"). Using
+         Release.Name for those releases yields peers that don't resolve and the HA
+         memberlist panics on startup. */}}
+  {{- $peer := printf "%s-%d.%s.%s.svc.cluster.local" $fullname $i $serviceName $.Release.Namespace }}
```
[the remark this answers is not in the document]
theTibi: Agreed, and the premise checks out: the only render assertions today are the node-exporter mode
cases, and `ct install` generates a `pmm-ha`-prefixed release name, so `pmm.fullname` equals
`Release.Name` there and this whole bug class is invisible to CI.

Added in ac33e22. One constraint worth recording: it can't live in the pure-render step —
`templates/pg-user-credentials-secrets.yaml:1-4` does an unconditional `lookup` + `fail` on
`pmm-secret`, which no `--set` disables and which `--show-only` doesn't dodge (Helm renders
everything before filtering). So it runs in the kind job after the `pmm-secret` prerequisite, as
`helm template pmm-2 charts/pmm-ha --namespace $NS --dry-run=server` (supported in the pinned Helm
3.15.4).

It asserts, for release `pmm-2`:

- `PMM_HA_PEERS` and the HAProxy init hostnames use `pmm-2-pmm-ha-<n>`, and the old `pmm-2-<n>`
  form is absent;
- the ClickHouse/Keeper scrape selectors are `regex: 'pmm-2'` / `'pmm-2-keeper'`;
- the CHI is still named from `Release.Name` — which is what makes those selectors correct, so if
  the CRs are ever renamed the check fails instead of the metrics silently disappearing;
- release `pmm.ha` renders `regex: 'pmm\.ha'`, so the `regexQuoteMeta` fix can't regress either.

Verified every assertion against a real render of the current chart, and confirmed no template is
`Capabilities.APIVersions`-gated, so rendering before the operators exist is safe.

PR #875, opened 2026-06-20
K8SPG-1057: Allow using etcd as patroni DCS

yoav-katz:
this PR adds the option to change dcs in the helm chart.


yoav-katz:
the code itself didnt pass all e2e tests yet so marking as draft :smile:

PR #879, opened 2026-06-24
K8SPS-627 add custom cluster name

nmarukovich:


charts/ps-db/templates/cluster.yaml
```diff
@@ -665,6 +665,9 @@ spec:
   {{- else }}
   {{- if $pmm.mysqlParams }}
     mysqlParams: {{ $pmm.mysqlParams }}
+  {{- end }}
+  {{- if $pmm.customClusterName }}
+      customClusterName: {{ $pmm.customClusterName }}
```
valmiranogueira: Seems like the indentation here isn't correct

charts/ps-db/values.yaml
```diff
@@ -681,6 +681,7 @@ pmm:
 #    failureThreshold: 3
   serverHost: monitoring-service
   # mysqlParams: "--disable-tablestats-limit=2000"
+  # customClusterName: "<string>"
```
valmiranogueira: I think we can have an example here

valmiranogueira (changes requested):

charts/ps-db/templates/cluster.yaml
```diff
@@ -665,6 +665,9 @@ spec:
...
   {{- end }}
```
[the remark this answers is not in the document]
valmiranogueira: Can we update the CRD also?

charts/ps-db/values.yaml
jvpasinatto: done

charts/ps-db/templates/cluster.yaml
jvpasinatto: done

charts/ps-db/templates/cluster.yaml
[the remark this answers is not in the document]
jvpasinatto: done

valmiranogueira (approved):

jvpasinatto (approved):

PR #882, opened 2026-06-26
K8SPS-127: automatic storage expansion

mayankshah1607:


charts/ps-db/templates/cluster.yaml
```diff
@@ -26,6 +26,10 @@ spec:
   crVersion: {{ .Chart.AppVersion }}
   {{- if .Values.enableVolumeExpansion }}
   enableVolumeExpansion: {{ .Values.enableVolumeExpansion }}
+  {{- end }}
+  {{- if .Values.storageScaling }}
+  storageScaling:
```
valmiranogueira, line 32: There was a discussion a while ago about our custom fields to be defined granularly in cluster template. This is based on https://perconadev.atlassian.net/browse/K8SPXC-1498 but it's supposed to be applied for all operators. Can we make this granular here then?

valmiranogueira (changes requested):

charts/ps-db/templates/cluster.yaml
mayankshah1607: yes we can, but I don't see how the benefit, it will just be a lot more verbose:

```yaml
  {{- $storageScaling := .Values.storageScaling }}
  storageScaling:
    {{- if $storageScaling.enableVolumeScaling }}
    enableVolumeScaling: {{ $storageScaling.enableVolumeScaling }}
    {{- end }}
    {{- if $storageScaling.enableExternalAutoscaling }}
    enableExternalAutoscaling: {{ $storageScaling.enableExternalAutoscaling }}
    {{- end }}
    {{- if $storageScaling.autoscaling }}
    {{- $autoscaling := $storageScaling.autoscaling }}
    autoscaling:
      {{- if $autoscaling.enabled }}
      enabled: {{ $autoscaling.enabled }}
      {{- end }}
      {{- if $autoscaling.triggerThresholdPercent }}
      triggerThresholdPercent: {{ $autoscaling.triggerThresholdPercent }}
      {{- end }}
      {{- if $autoscaling.growthStep }}
      growthStep: {{ $autoscaling.growthStep }}
      {{- end }}
      {{- if $autoscaling.maxSize }}
      maxSize: {{ $autoscaling.maxSize }}
      {{- end }}
    {{- end }}
  {{- end }}
```

Is this what we want?

valmiranogueira: Yes, this is what was agreed to solve the above Jira, old items will be changed to granular (only custom ones, not k8s like affinity).

mayankshah1607: https://github.com/percona/percona-helm-charts/pull/882/changes/0567a7fa65c3bc59b4b471b14529ad47bdb9b9f2

charts/ps-db/values.yaml
```diff
@@ -17,7 +17,16 @@ finalizers:
 #    proxy: false
 #    proxySize: false
 crVersion: 1.2.0
+# enableVolumeExpansion is deprecated, use storageScaling.enableVolumeScaling instead.
```
valmiranogueira: Sorry last comment, add deprecation to readme, example:
`| ~~`haproxy.livenessDelaySec`~~                    | **(Deprecated since 1.19.0)** HAProxy Pod delay for liveness probe in seconds    `
The value can be removed from here

valmiranogueira (changes requested):

eleo007 (approved):

valmiranogueira (approved):

jvpasinatto (approved):

PR #888, opened 2026-07-01
update crds

mayankshah1607:


DhruthiKV:
Reverted the ClusterSync CRD changes because they are already part of PR #872. Including them here caused a large unrelated CRD diff and made the CRD sync check fail. This PR now contains only the CRD changes relevant to K8SPSMDB-1547.

jvpasinatto (approved):

PR #892, opened 2026-07-03
[psmdb-db] add containerSecurityContext to logcollector, too.

### LINKED ISSUE #891: [psmd-db] logcollector containerSecurityContext missing
The CRD of `PerconaServerMongoDB` has the field `.spec.logcollector.containerSecurityContext` which is missing in the helm config of the CRD.

belfhi:
Fixes #891


charts/psmdb-db/values.yaml
```diff
@@ -869,3 +869,4 @@ logcollector:
     requests:
       memory: 100M
       cpu: 200m
+  containerSecurityContext: {}
```
jvpasinatto: please comment out this line as the other occurrences of  containerSecurityContext

charts/psmdb-db/templates/cluster.yaml
```diff
@@ -918,3 +918,8 @@ spec:
     resources:
 {{ .Values.logcollector.resources | toYaml | indent 6 }}
     {{- end }}
+    {{- if .Values.logcollector.containerSecurityContext }}
+    containerSecurityContext:
+{{- .Values.logcollector.containerSecurityContext | toYaml | nindent 6 }}
```
jvpasinatto: please, remove dash from `{{- .Values.logcollector` use the other containerSecurityContext as reference

charts/psmdb-db/templates/cluster.yaml
```diff
@@ -918,3 +918,8 @@ spec:
...
+    {{- end }}
+
```
jvpasinatto, line 1027: please, add logcollector.containerSecurityContext field in readme as well

jvpasinatto (changes requested):

charts/psmdb-db/values.yaml
```diff
@@ -860,12 +860,13 @@ logcollector:
...
-#  configuration: |
-#    [SERVICE]
-#        Flush        1
-#        Log_Level    error
-#        Daemon       off
+  #  configuration: |
+  #    [SERVICE]
+  #        Flush        1
+  #        Log_Level    error
+  #        Daemon       off
```
egegunes, line 1067: @belfhi please restore the old version of comments. by convention we put the comment character at the beginning of the line and indentation should be correct when you remove `#` character

egegunes (changes requested):

charts/psmdb-db/values.yaml
belfhi: but if you look at the values.yaml this is clearly not the case and from the current state of the file I thought the convention was to have the comment start at the intended indentation level. I can change it back but my change is more coherent with the rest of file imho ;)

egegunes: yes, looks like you're right. this is different our convention in the operator repo

egegunes (dismissed):

jvpasinatto (approved):

jvpasinatto:
I moved to release 1.23.0 branch and i will solve some conflict errors there.
Thanks for the contribution @belfhi !

Issue #825, opened 2026-03-31
Add option to CRD helm chart to protect CRDs from being deleted


horihel:
Helm allows to set resources so they won't be deleted on uninstall: https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource

I'm not the first one that saw flux (or any other gitops tool) accidentally uninstall/reinstall a chart because of some dumb mistake. In case of psmdb-operator CRD, this could be devastating.

Adding the aforementioned annotation to the CRD helm charts (even optionally enabled through values.yaml) could be a safety net that prevents downtime or data-loss.

PR #901, opened 2026-07-10
K8SPSMDB-1633 - protect CRDs from deletion via preserveCrds

DrFaust92:
## Summary

Resolves the psmdb part of #825 ([K8SPSMDB-1633](https://perconadev.atlassian.net/browse/K8SPSMDB-1633)).

Adds an **opt-in** `preserveCrds` value to the `psmdb-operator-crds` chart. When enabled, the [`helm.sh/resource-policy: keep`](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource) annotation is stamped onto all three CRDs, so a `helm uninstall` (or a GitOps tool like Flux/Argo CD accidentally pruning the release) leaves the CRDs — and every cluster stored under them — in place instead of deleting them.

Since these CRDs live in `templates/` (so they can be upgraded), they are otherwise subject to deletion on uninstall — which, as the issue notes, can be devastating.

## Changes

- `values.yaml`: new `preserveCrds: false` (documented; disabled by default → no behavior change for existing users).
- CRD templates: conditionally inject `helm.sh/resource-policy: keep` into `metadata.annotations`.
- `README.md`: new "Protecting CRDs from Deletion" section + values table.
- `psmdb-operator-crd-sync-check.yaml`: the content-sync check strips the templated annotation before diffing, so the CRD content is still validated against `psmdb-operator/crds/crd.yaml`.

## Testing

- `helm template` (default) → CRDs render identically to before (0 annotations).
- `helm template --set preserveCrds=true` → all 3 CRDs get the annotation.
- CRD content-sync diff → in sync ✓
- `helm lint` → 0 failures.

## Notes

Kept intentionally opt-in (`false`) to avoid changing uninstall behavior for existing users. Happy to default it to `true` if maintainers prefer safe-by-default. The sibling tickets (K8SPXC-1849 / K8SPS-677 / K8SPG-996) can follow the same pattern in separate PRs.

[K8SPSMDB-1633]: https://perconadev.atlassian.net/browse/K8SPSMDB-1633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

### ⚠️ Versioning note for maintainers

This is the first standalone change to the `psmdb-operator-crds` chart, which surfaced a conflict between two CI gates:

- `ct lint` requires a chart **version bump** on any chart change.
- `psmdb-operator-crd-sync-check` pinned the crds chart `version` to equal the operator `appVersion`.

Together these made a standalone crds-chart change impossible to land. I resolved it by:

- Bumping `psmdb-operator-crds` chart `version` `1.22.0 → 1.22.1` (appVersion stays `1.22.0`, pinned to the operator).
- Relaxing the sync-check so the crds chart `version` only needs to share the operator `appVersion`'s **major.minor** and may carry its own patch bump — mirroring how the operator chart version already floats above its appVersion. `appVersion` remains strictly pinned, and CRD **content** is still validated byte-for-byte by the existing content-sync diff.

Happy to adjust if you'd prefer a different versioning convention here.

### DIFF
--- a/charts/psmdb-operator-crds/values.yaml
+++ b/charts/psmdb-operator-crds/values.yaml
@@ -1,2 +1,9 @@
 # Default values for psmdb-operator-crds.
-# This chart contains CRDs only and does not require any configuration values.
+# This chart contains CRDs only; all values below are optional.
+
+# preserveCrds adds the "helm.sh/resource-policy: keep" annotation to the CRDs so
+# they (and every custom resource / cluster stored under them) survive a
+# `helm uninstall` of this chart. Set to true to guard against a GitOps tool
+# (Flux, Argo CD, etc.) accidentally deleting the CRDs. See:
+# https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource
+preserveCrds: false
--- a/charts/psmdb-operator-crds/values.yaml
+++ b/charts/psmdb-operator-crds/values.yaml
@@ -3,7 +3,8 @@
 
 # preserveCrds adds the "helm.sh/resource-policy: keep" annotation to the CRDs so
 # they (and every custom resource / cluster stored under them) survive a
-# `helm uninstall` of this chart. Set to true to guard against a GitOps tool
-# (Flux, Argo CD, etc.) accidentally deleting the CRDs. See:
+# `helm uninstall` of this chart -- a safety net against a GitOps tool
+# (Flux, Argo CD, etc.) accidentally deleting the CRDs and destroying data.
+# Enabled by default; set to false to let uninstall remove the CRDs. See:
 # https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource
-preserveCrds: false
+preserveCrds: true
--- a/.github/workflows/psmdb-operator-crd-sync-check.yaml
+++ b/.github/workflows/psmdb-operator-crd-sync-check.yaml
@@ -38,11 +38,16 @@ jobs:
           
           VERSION_SYNC=true
           
-          # psmdb-operator-crds chart version should match operator appVersion (CRDs are tied to operator version, not chart version)
+          # psmdb-operator-crds chart version must share the operator appVersion's
+          # major.minor (CRDs are tied to the operator's release line) but may carry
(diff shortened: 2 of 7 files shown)


charts/psmdb-operator-crds/values.yaml
```diff
@@ -1,2 +1,9 @@
 # Default values for psmdb-operator-crds.
-# This chart contains CRDs only and does not require any configuration values.
+# This chart contains CRDs only; all values below are optional.
+
+# preserveCrds adds the "helm.sh/resource-policy: keep" annotation to the CRDs so
+# they (and every custom resource / cluster stored under them) survive a
+# `helm uninstall` of this chart. Set to true to guard against a GitOps tool
+# (Flux, Argo CD, etc.) accidentally deleting the CRDs. See:
+# https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource
+preserveCrds: false
```
egegunes: i wonder if this should be true by default

DrFaust92: Agreed -- I've flipped the default to `true` (89c5871).

The failure modes are asymmetric: with `false`, an accidental `helm uninstall` / GitOps prune cascades through the CRDs to every `PerconaServerMongoDB` and its data (irreversible). With `true`, the worst case is orphaned, cluster-scoped CRDs left after a teardown (recoverable with `kubectl delete crd`). Safe-by-default is also the ecosystem norm here (cert-manager, prometheus-operator both keep CRDs). And since `helm.sh/resource-policy: keep` only affects delete/uninstall -- not upgrade -- flipping the default doesn't touch the templated-CRD upgrade path these live in `templates/` for.

One caveat worth flagging: it's a behavior change on uninstall, so ephemeral/CI flows that uninstall+reinstall expecting a clean cluster will now leave CRDs behind. I documented the opt-out (`preserveCrds: false`) in the README/values for that case.

I'll make the PXC/PS/PG siblings match once the pattern here is settled so all four charts stay consistent.

valmiranogueira (approved):

eleo007 (approved):

PR #916, opened 2026-07-29
PMM-15070 Scrape OpenShift node exporter

4nte:
[PMM-15070](https://perconadev.atlassian.net/browse/PMM-15070)

There are two main issues resolved with this PR:
1. **On OpenShift, the chart's bundled node-exporter can't run.** It uses host networking and
   binds port 9100, which OpenShift's own platform node-exporter already holds, so the DaemonSet
   crash-loops on "address already in use".

2. **Cluster nodes never showed up in dashboards.** The node-exporter scrape job only
   labelled metrics with `node`, `namespace` and `pod`. PMM's OS dashboards (Node Summary, Nodes
   Overview, Nodes Compare, Disk Details) pick nodes from the `node_name` label, so the nodes were
   collected but invisible. The "Node Name" filter only listed the pods that run a pmm-agent (the PMM Server pods and the PG instances).
 

[PMM-15070]: https://perconadev.atlassian.net/browse/PMM-15070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JiriCtvrtka:
Chart version not bumped. [charts/pmm-ha/Chart.yaml](https://github.com/percona/percona-helm-charts/blob/781cf5f5/charts/pmm-ha/Chart.yaml) is still version: 1.6.0, identical to main, and pmm-ha-1.6.0 is already published (it's in the live index.yaml). release.yaml runs helm/chart-releaser-action with skip_existing: true, so merging as-is silently publishes nothing — the fix never reaches users. CONTRIBUTING also requires a patch bump per chart change. Bump to 1.6.1. (The --check-version-increment=true in pmm-ha-pr-checks.yaml did not catch this — likely the sparse checkout defeats ct's target-branch comparison. Worth a follow-up issue.)

@4nte is this valid one or we have different workflow?

ademidoff:
> Chart version not bumped. [charts/pmm-ha/Chart.yaml](https://github.com/percona/percona-helm-charts/blob/781cf5f5/charts/pmm-ha/Chart.yaml) is still version: 1.6.0, identical to main, and pmm-ha-1.6.0 is already published (it's in the live index.yaml). release.yaml runs helm/chart-releaser-action with skip_existing: true, so merging as-is silently publishes nothing — the fix never reaches users. CONTRIBUTING also requires a patch bump per chart change. Bump to 1.6.1. (The --check-version-increment=true in pmm-ha-pr-checks.yaml did not catch this — likely the sparse checkout defeats ct's target-branch comparison. Worth a follow-up issue.)
> 
> @4nte is this valid one or we have different workflow?

No, we bump up the version at the time of relaease, so we can merge PR without releasing.

ademidoff (commented):
## Review

### What it does

Adds a `nodeExporter.mode` switch (`internal` | `openshift`) to `charts/pmm-ha`. In `openshift` mode the VMAgent `inlineScrapeConfig` drops the bundled `prometheus-node-exporter` job and instead scrapes OpenShift's platform node-exporter in `openshift-monitoring` over HTTPS through its kube-rbac-proxy (SA token + service-CA verification with `server_name` pinned). Both jobs now share a relabel helper that adds `node_name` and `node_type=generic` so nodes appear in PMM's OS dashboards. A `fail()` helper enforces `prometheus-node-exporter.enabled=false` when mode is `openshift`, and NOTES.txt echoes the active mode.


### Should fix

- **Validation is bypassable and the bundled DaemonSet still deploys.** `charts/pmm-ha/templates/_helpers.tpl:168` uses `index .Values "prometheus-node-exporter"`; if a user sets `prometheus-node-exporter: null` (values file or `--set`), the key disappears, `$sub` is nil, validation passes — but Helm falls back to the dependency's own default and still renders `prometheus-node-exporter/templates/daemonset.yaml`, i.e. exactly the port-9100 collision the check exists to prevent. Verified locally. The same nil makes `charts/pmm-ha/templates/NOTES.txt:59` print a false "no node metrics will be collected" warning in internal mode. Suggest one source of truth with a default:

  ```
  {{- $bundled := dig "prometheus-node-exporter" "enabled" true .Values -}}
  ```

  consumed by both the validate helper and NOTES.

- **`nodeExporter: null` produces a raw template error instead of the friendly `fail`.** With `--set nodeExporter=null` the failure surfaces as `nil pointer evaluating interface {}.mode` at `charts/pmm-ha/templates/vmagent.yaml:249` — vmagent.yaml is evaluated *before* statefulset.yaml, so the "fail-fast … called from statefulset.yaml" comment at `charts/pmm-ha/templates/_helpers.tpl:159-160` doesn't hold for ordering. A `pmm.nodeExporter.mode` helper (`dig "nodeExporter" "mode" "internal" .Values`) used by validate, vmagent.yaml and NOTES.txt would cover this.

- **No CI coverage of the new branch.** `.github/workflows/pmm-ha-pr-checks.yaml` lints and installs with default values only, so `mode=openshift` is never rendered. The cheapest guard is a render step (server-side dry run is required because `charts/pmm-ha/templates/pg-user-credentials-secrets.yaml` does a `lookup`):

  ```sh
  helm install smoke charts/pmm-ha -n "$NS" --dry-run=server \
    --set nodeExporter.mode=openshift --set prometheus-node-exporter.enabled=false \
    | grep -q "job_name: 'openshift-node-exporter'"
  ```

  Adding a `ci/*-values.yaml` for this is not a good fit — `ct` would attempt a real install in kind and the job has no targets there.

### Nits

- `charts/pmm-ha/templates/vmagent.yaml:265`: prefer `{{- else if eq .Values.nodeExporter.mode "openshift" }}` so an unvalidated mode can't silently select the OpenShift job.
- `charts/pmm-ha/templates/_helpers.tpl:214` and `charts/pmm-ha/templates/_helpers.tpl:217`: `regex: node-exporter` / `regex: https` are unquoted; every other regex in the inline config is single-quoted.
- `charts/pmm-ha/README.md:320`: the heading "Node exporter source parameters" doesn't match the `@section` text in `charts/pmm-ha/values.yaml:696`, so a future README regeneration will churn; the table separator row widths also don't line up with the header (cosmetic only).
- `charts/pmm-ha/README.md:551`: the cpufreq snippet doesn't say it's the `cluster-monitoring-config` ConfigMap in `openshift-monitoring` (and that it may not exist yet). One line would save a support round-trip.
- Worth a doc line: `mode=openshift` relies on the VM operator running cluster-wide. If the operator is scoped with `WATCH_NAMESPACE`, it generates a namespaced Role and the `openshift-monitoring` SD returns zero targets with no error.

### Other risks to call out (not defects)

- **`mode=openshift` on a non-OpenShift cluster fails quietly.** The job renders, then each scrape errors with `cannot read ca_file` (VM reads `ca_file` lazily per connection, so only this job dies, not the whole config). A `.Capabilities.APIVersions.Has "security.openshift.io/v1"` guard would fail fast, but it also breaks plain `helm template` without `--api-versions`, so a documented prerequisite is probably the better trade.
- **Collector-set differences beyond cpufreq.** OpenShift's node-exporter also differs in excluded filesystems/mountpoints and enabled collectors. If other empty panels showed up during testing, listing them in the Limitations block would help; otherwise it's worth stating explicitly that cpufreq was the only gap observed.

ademidoff (approved):

JiriCtvrtka (approved):

PR #925, opened 2026-08-10
PMM-15309 Add read-only ClickHouse data source user

ademidoff:
## What

Gives the Grafana ClickHouse data source its own read-only ClickHouse user, separate from the account PMM writes Query Analytics data with.

Chart side of [percona/pmm#5747](https://github.com/percona/pmm/pull/5747), which lands in PMM 3.9.1. Only `pmm-ha` needs changes — the `pmm` chart has no ClickHouse configuration and keeps using the built-in server, whose defaults ship with PMM Server.

## Why

Grafana executes data source queries on behalf of every signed-in user, including Viewers. Today the data source connects as `PMM_CLICKHOUSE_USER`, which the chart creates with `access_management: "1"` and the default profile, so any Viewer can run arbitrary SQL with that account's privileges — `INSERT`, `ALTER`, `DROP`, and the `SOURCES` privileges that reach table functions such as `url()` and `file()`.

PMM 3.9.1 reads the data source credentials from `PMM_CLICKHOUSE_DATASOURCE_USER` and `PMM_CLICKHOUSE_DATASOURCE_PASSWORD`. Without this change the data source falls back to defaults that do not exist on the Altinity cluster, and every ClickHouse-backed panel stops working after upgrade.

## How

- `users.d/pmm-datasource-user.xml` drop-in creates the user with `GRANT SELECT` on the Query Analytics database and nothing else. A drop-in rather than the operator's `users:` map because that map cannot express `<grants>`, and withholding the `SOURCES` privileges is what blocks the table functions.
- Profile uses `readonly=1` with `max_execution_time` marked `changeable_in_readonly`, per the [Grafana plugin guidance](https://grafana.com/docs/plugins/grafana-clickhouse-datasource/latest/configure/#clickhouse-user-and-permissions). The plugin sets that setting on every query, and `readonly=2` is discouraged for shared instances.
- `config.d/pmm-settings-constraints.xml` enables `settings_constraints_replace_previous`, without which ClickHouse rejects `changeable_in_readonly` with `NOT_IMPLEMENTED`.
- Credentials live in a chart-managed secret, `<release>-pmm-clickhouse-datasource`, generated on first install and read back on upgrade.

## No action required on upgrade

`pmm-secret` is untouched. Since the chart creates the ClickHouse user itself, it picks the password too, so nothing has to be added to a hand-created secret and existing deployments upgrade unchanged. `clickhouse.datasource.user` and `clickhouse.datasource.password` are available for anyone who wants to pin them.

## Testing

`helm template` and `helm lint` pass, and the generated password in the secret matches the SHA-256 in the ClickHouse drop-in, verified for three configurations:

| Configuration | Result |
|---|---|
| defaults (`secret.create: false`) | password generated, hash matches |
| `secret.create: true` | password generated, hash matches |
| `clickhouse.datasource.password` pinned | override honoured, hash matches |

The password is memoised on `.Values`, since `randAlphaNum` would otherwise hand a different value to the secret and to the drop-in and leave Grafana unable to authenticate. Both embedded XML documents validate with `xmllint`.

Not yet exercised against a live cluster, and worth confirming before merge:

- that the Altinity operator accepts a `config.d/` key under `files:` — the chart only used `users.d/` until now;
- that ClickHouse merge order places this drop-in after the operator's generated users file.

Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4516

### DIFF
--- a/charts/pmm-ha/templates/_helpers.tpl
+++ b/charts/pmm-ha/templates/_helpers.tpl
@@ -232,19 +232,24 @@ when nodeExporter.mode == "openshift".
 {{- end -}}
 
 {{/*
+Name of the chart-managed secret holding the read-only ClickHouse data source credentials.
+
+Kept apart from .Values.secret.name because that secret is user-managed by default, and these
+credentials are internal: the chart creates the ClickHouse user itself, so nobody has to supply
+them.
+*/}}
+{{- define "pmm.clickhouse.datasourceSecretName" -}}
+{{- printf "%s-clickhouse-datasource" (include "pmm.fullname" .) -}}
+{{- end -}}
+
+{{/*
 Username of the read-only ClickHouse user backing the Grafana data source.
 
 Grafana runs data source queries on behalf of every signed-in user, including Viewers, so this
-must not be the user PMM writes Query Analytics data with. A username already stored in the
-secret wins, so that upgrades keep using the account that exists in ClickHouse.
+must not be the user PMM writes Query Analytics data with.
 */}}
 {{- define "pmm.clickhouse.datasourceUser" -}}
-{{- $existing := (lookup "v1" "Secret" .Release.Namespace .Values.secret.name) -}}
-{{- if and $existing (index $existing.data "PMM_CLICKHOUSE_DATASOURCE_USER") -}}
-{{- index $existing.data "PMM_CLICKHOUSE_DATASOURCE_USER" | b64dec -}}
-{{- else -}}
-{{- .Values.secret.clickhouse_datasource_user | default "clickhouse_pmm_readonly" -}}
-{{- end -}}
+{{- .Values.clickhouse.datasource.user | default "clickhouse_pmm_readonly" -}}
 {{- end -}}
 
 {{/*
@@ -253,15 +258,14 @@ Password of the read-only ClickHouse data source user.
 The secret hands the plaintext to PMM while the ClickHouse drop-in needs its SHA-256, so both
 have to agree. A generated password is therefore memoised on .Values: randAlphaNum would
 otherwise return a different value to each caller and leave Grafana unable to authenticate.
+Once generated it is read back from the chart-managed secret, so upgrades keep the same value.
 */}}
 {{- define "pmm.clickhouse.datasourcePassword" -}}
-{{- $existing := (lookup "v1" "Secret" .Release.Namespace .Values.secret.name) -}}
-{{- if and $existing (index $existing.data "PMM_CLICKHOUSE_DATASOURCE_PASSWORD") -}}
+{{- $existing := (lookup "v1" "Secret" .Release.Namespace (include "pmm.clickhouse.datasourceSecretName" .)) -}}
+{{- if .Values.clickhouse.datasource.password -}}
+{{- .Values.clickhouse.datasource.password -}}
+{{- else if and $existing (index $existing.data "PMM_CLICKHOUSE_DATASOURCE_PASSWORD") -}}
 {{- index $existing.data "PMM_CLICKHOUSE_DATASOURCE_PASSWORD" | b64dec -}}
-{{- else if .Values.secret.clickhouse_datasource_password -}}
-{{- .Values.secret.clickhouse_datasource_password -}}
-{{- else if and $existing (not .Values.secret.create) -}}
-{{- fail (printf "Secret '%s' is missing required key 'PMM_CLICKHOUSE_DATASOURCE_PASSWORD'. The Grafana ClickHouse data source now connects as a read-only user, separate from PMM_CLICKHOUSE_USER. Add PMM_CLICKHOUSE_DATASOURCE_USER and PMM_CLICKHOUSE_DATASOURCE_PASSWORD to the secret, or set secret.clickhouse_datasource_password." .Values.secret.name) -}}
 {{- else -}}
 {{- if not (hasKey .Values "generatedClickhouseDatasourcePassword") -}}
 {{- $_ := set .Values "generatedClickhouseDatasourcePassword" (randAlphaNum 32) -}}
--- a/charts/pmm-ha/templates/_helpers.tpl
+++ b/charts/pmm-ha/templates/_helpers.tpl
@@ -230,3 +230,42 @@ when nodeExporter.mode == "openshift".
       action: keep
     {{- include "pmm.nodeExporter.pmmRelabelConfigs" . | nindent 4 }}
 {{- end -}}
+
+{{/*
+Username of the read-only ClickHouse user backing the Grafana data source.
+
+Grafana runs data source queries on behalf of every signed-in user, including Viewers, so this
+must not be the user PMM writes Query Analytics data with. A username already stored in the
+secret wins, so that upgrades keep using the account that exists in ClickHouse.
+*/}}
+{{- define "pmm.clickhouse.datasourceUser" -}}
+{{- $existing := (lookup "v1" "Secret" .Release.Namespace .Values.secret.name) -}}
+{{- if and $existing (index $existing.data "PMM_CLICKHOUSE_DATASOURCE_USER") -}}
+{{- index $existing.data "PMM_CLICKHOUSE_DATASOURCE_USER" | b64dec -}}
+{{- else -}}
+{{- .Values.secret.clickhouse_datasource_user | default "clickhouse_pmm_readonly" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Password of the read-only ClickHouse data source user.
+
+The secret hands the plaintext to PMM while the ClickHouse drop-in needs its SHA-256, so both
+have to agree. A generated password is therefore memoised on .Values: randAlphaNum would
+otherwise return a different value to each caller and leave Grafana unable to authenticate.
+*/}}
+{{- define "pmm.clickhouse.datasourcePassword" -}}
+{{- $existing := (lookup "v1" "Secret" .Release.Namespace .Values.secret.name) -}}
+{{- if and $existing (index $existing.data "PMM_CLICKHOUSE_DATASOURCE_PASSWORD") -}}
+{{- index $existing.data "PMM_CLICKHOUSE_DATASOURCE_PASSWORD" | b64dec -}}
+{{- else if .Values.secret.clickhouse_datasource_password -}}
+{{- .Values.secret.clickhouse_datasource_password -}}
+{{- else if and $existing (not .Values.secret.create) -}}
+{{- fail (printf "Secret '%s' is missing required key 'PMM_CLICKHOUSE_DATASOURCE_PASSWORD'. The Grafana ClickHouse data source now connects as a read-only user, separate from PMM_CLICKHOUSE_USER. Add PMM_CLICKHOUSE_DATASOURCE_USER and PMM_CLICKHOUSE_DATASOURCE_PASSWORD to the secret, or set secret.clickhouse_datasource_password." .Values.secret.name) -}}
+{{- else -}}
+{{- if not (hasKey .Values "generatedClickhouseDatasourcePassword") -}}
+{{- $_ := set .Values "generatedClickhouseDatasourcePassword" (randAlphaNum 32) -}}
+{{- end -}}
+{{- get .Values "generatedClickhouseDatasourcePassword" -}}
+{{- end -}}
+{{- end -}}
--- a/charts/pmm-ha/templates/_helpers.tpl
+++ b/charts/pmm-ha/templates/_helpers.tpl
@@ -73,6 +73,12 @@ Pod annotation
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 helm.sh/chart: {{ include "pmm.chart" . }}
 checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+{{/*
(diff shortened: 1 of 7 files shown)


charts/pmm-ha/values.yaml
```diff
@@ -11,7 +11,7 @@
 ## @param image.imagePullSecrets Global Docker registry secret names as an array
 ##
 image:
-  repository: percona/pmm-server
+  repository: perconalab/pmm-server-fb
```
theTibi: this should be `percona/pmm-server`

charts/pmm-ha/templates/clickhouse-cluster.yaml
```diff
@@ -80,6 +83,56 @@ spec:
...
+        </clickhouse>
+
+      # readonly=1 blocks every setting change, but the Grafana ClickHouse plugin sets
+      # max_execution_time on each query to enforce its timeout. Marking that one setting
+      # changeable needs this flag, without which ClickHouse rejects the constraint outright.
+      config.d/pmm-settings-constraints.xml: |
+        <clickhouse>
+          <access_control_improvements>
+            <settings_constraints_replace_previous>true</settings_constraints_replace_previous>
+          </access_control_improvements>
```
theTibi, line 152: I think tis should be:
```
<clickhouse>
  <access_control_improvements>
    <settings_constraints_replace_previous>true</settings_constraints_replace_previous>
  </access_control_improvements>
  <remote_url_allow_hosts></remote_url_allow_hosts>   <!-- add: parity with pmm#5747 -->
</clickhouse>
```
which blocks URL-engine for every user

charts/pmm-ha/values.yaml
ademidoff: Reverted.

charts/pmm-ha/templates/clickhouse-cluster.yaml
```diff
@@ -80,6 +83,56 @@ spec:
...
+              <networks>
+                {{- range .Values.clickhouse.users.networks.ip }}
+                <ip>{{ . }}</ip>
+                {{- end }}
+              </networks>
+              <profile>pmm_datasource</profile>
+              <quota>default</quota>
+              <grants>
+                <query>GRANT SELECT ON {{ $clickhouseDatabase }}.*</query>
+              </grants>
```
4nte, line 126: The <grants> allowlist does not deliver its stated invariant - the datasource user can read all of `system.*`
`default-users.xml:115-118` states the goal ("must not be able to do more than read QAN data. The `<grants>` section is what enforces that"). It does not, because of two pre-existing settings the PR leaves untouched, which I verified are present in all three config files:

default-config.xml:675      `<select_from_system_db_requires_grant>false</...>`
default-config.xml:680      `<select_from_information_schema_requires_grant>false</...>`

In that mode ClickHouse grants SELECT on the whole system database to every user regardless of <grants>. Reproduced on the pinned 25.3.6.56 as grafana/grafana: system.query_log readable (returned the full text of a CREATE USER ... IDENTIFIED WITH sha256_password ... statement), plus system.text_log and system.processes. Both logs are uncommented in the shipped config (:938, :1053), so these tables exist on a real server. Under the PR's own threat model - Grafana runs datasource queries on behalf of any signed-in user - an Explore session can read the SQL text of every query any user has run.

Not a regression, but it is the exact property the PR sets out to establish, so it belongs in scope. Flipping `select_from_system_db_requires_grant` to `true` was verified safe against the full plugin metadata surface (system.tables/columns/databases/settings/functions/one, SHOW DATABASES, DESC TABLE, version()); only query_log, text_log, clusters, build_options become denied. qan-api2 is unaffected (default holds GRANT ALL).

charts/pmm-ha/templates/clickhouse-cluster.yaml
```diff
@@ -5,6 +5,9 @@
 {{- else if .Values.secret.clickhouse_user }}
 {{- $clickhouseUser = .Values.secret.clickhouse_user }}
 {{- end }}
+{{- $datasourceUser := include "pmm.clickhouse.datasourceUser" . }}
```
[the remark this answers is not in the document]
ademidoff: Good catch — fixed in 9b33502a. Rendering now fails when `clickhouse.datasource.user` equals the PMM writer or `clickhouse_operator`, since a drop-in keyed on the same username would have rewritten that account's profile and grants rather than creating a separate one.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -230,3 +230,46 @@ when nodeExporter.mode == "openshift".
...
+
+{{/*
+Username of the read-only ClickHouse user backing the Grafana data source.
+
+Grafana runs data source queries on behalf of every signed-in user, including Viewers, so this
+must not be the user PMM writes Query Analytics data with.
+*/}}
+{{- define "pmm.clickhouse.datasourceUser" -}}
+{{- .Values.clickhouse.datasource.user | default "clickhouse_pmm_readonly" -}}
+{{- end -}}
```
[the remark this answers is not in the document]
ademidoff: Fixed in 9b33502a. The username is validated against `^[A-Za-z_][A-Za-z0-9_-]*$` before it is written, so `123-reader` is now rejected at render time instead of producing a users file ClickHouse cannot load.

charts/pmm-ha/templates/clickhouse-cluster.yaml
```diff
@@ -80,6 +83,56 @@ spec:
...
+              <password_sha256_hex>{{ $datasourcePassword | sha256sum }}</password_sha256_hex>
+              <networks>
+                {{- range .Values.clickhouse.users.networks.ip }}
+                <ip>{{ . }}</ip>
+                {{- end }}
+              </networks>
+              <profile>pmm_datasource</profile>
+              <quota>default</quota>
+              <grants>
+                <query>GRANT SELECT ON {{ $clickhouseDatabase }}.*</query>
```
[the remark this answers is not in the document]
ademidoff: Fixed in 9b33502a, in both respects. The database name is validated with the same pattern, which rules out `&` and anything else that would break the XML, and it is now backquoted in the GRANT so a name such as `my-pmm` stays valid SQL. Verified: `--set pmmEnv.PMM_CLICKHOUSE_DATABASE=my-pmm` renders `GRANT SELECT ON ` + "`my-pmm`" + `.*`, and `a&b` fails the render.

charts/pmm-ha/templates/statefulset.yaml
```diff
@@ -229,6 +229,18 @@ spec:
                 secretKeyRef:
                   name: {{ .Values.secret.name }}
                   key: PMM_CLICKHOUSE_PASSWORD
+            # Read-only credentials for the Grafana ClickHouse data source, kept apart from the
+            # privileged pair above because Grafana queries run on behalf of every signed-in user.
+            - name: PMM_CLICKHOUSE_DATASOURCE_USER
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "pmm.clickhouse.datasourceSecretName" . }}
+                  key: PMM_CLICKHOUSE_DATASOURCE_USER
```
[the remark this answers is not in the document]
ademidoff: Fixed in 9b33502a — a `checksum/clickhouse-datasource` annotation over the credentials secret now sits in the pod template, alongside the existing `checksum/config`.

Verified both properties, since a checksum that changes on every render would cause a rollout on every upgrade: with the password pinned, two renders produce the same value; changing it produces a different one. The generated password is read back from the secret on upgrade, so it is stable in practice.

charts/pmm-ha/values.yaml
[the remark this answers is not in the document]
ademidoff: Already addressed in a82ac5cf, before this review landed — the chart default is back to `percona/pmm-server:3.9.0`. The feature-build image is applied as a local override for testing, not committed. Agreed on the tag: it needs bumping to 3.9.1 once that release exists, since that is where the new environment variables ship.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -230,3 +230,46 @@ when nodeExporter.mode == "openshift".
...
+
+{{/*
+Name of the chart-managed secret holding the read-only ClickHouse data source credentials.
+
+Kept apart from .Values.secret.name because that secret is user-managed by default, and these
+credentials are internal: the chart creates the ClickHouse user itself, so nobody has to supply
+them.
+*/}}
+{{- define "pmm.clickhouse.datasourceSecretName" -}}
+{{- printf "%s-clickhouse-datasource" (include "pmm.fullname" .) -}}
```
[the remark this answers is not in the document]
ademidoff: Correct, and the docs were the wrong side of it — fixed in 9b33502a. The name follows the chart fullname, so the documented `helm install pmm-ha` release yields `pmm-ha-clickhouse-datasource`. Confirmed by rendering three release names: `pmm-ha` gives `pmm-ha-clickhouse-datasource`, `pmm` gives `pmm-pmm-ha-clickhouse-datasource`, `monitoring` gives `monitoring-pmm-ha-clickhouse-datasource`. The README and values.yaml now use the documented name and note that it tracks the release name.

charts/pmm-ha/templates/clickhouse-cluster.yaml
ademidoff: Agreed, added in c6ede21b — the `config.d` drop-in (now `pmm-hardening.xml`) carries `remote_url_allow_hosts` alongside the constraints flag.

I checked the one thing that could have made this unsafe on a real cluster, since an empty allowlist also blocks `remote()`, not just `url()`: Query Analytics never uses either. Replication runs over the interserver protocol and Keeper, and schema changes go through `ON CLUSTER` DDL on a `Replicated` database — `qan-api2/db.go:132`. There are no `Distributed` tables and no `remote()` calls anywhere in qan-api2.

One residual risk worth exercising during cluster testing: the Altinity operator propagates schema between hosts when replicas are added. I could not confirm from here whether it does that through its own connections or through `remote()`. If it is the latter, scaling replicas would surface it.

charts/pmm-ha/templates/clickhouse-cluster.yaml
ademidoff: Thanks — the mechanism is real, and I reproduced it. But the location matters, and it points more at percona/pmm#5747 than at this chart.

Reproduced on stock `clickhouse/clickhouse-server:25.3.6.56` with this PR's exact drop-in (readonly=1 profile, `GRANT SELECT ON pmm.*`):

| `select_from_system_db_requires_grant` | `system.query_log` as the data source user |
|---|---|
| `false` | readable — 36 rows, and `SELECT substring(query,1,45) ...` returned `CREATE DATABASE IF NOT EXISTS pmm` |
| `true` | `Code: 497 ... necessary to have the grant SELECT for at least one column on system.query_log` |

So your read is right: with the flag off, `<grants>` does not deliver the stated invariant.

Two things I did not expect:

1. **Stock ClickHouse 25.3.6.56 already ships the flag as `true`** — `/etc/clickhouse-server/config.xml:789` in that image. The value `false` you found comes from PMM's forked config, which carries an older upstream default. The Altinity image this chart runs is likely `true` as well, so the HA cluster was probably not exposed. I have set it explicitly in c6ede21b anyway, so the guarantee no longer depends on the image's default.
2. **The exposed deployment is the built-in ClickHouse**, where `default-config.xml`, `low-memory-config.xml` and `dev/clickhouse-config.xml` all set it to `false`. That is percona/pmm#5747, not this PR.

I also confirmed your compatibility claim on the same container: with the flag on, `system.tables`, `system.columns`, `system.databases`, `SHOW DATABASES`, `DESCRIBE TABLE` and `version()` all still work for the restricted user; `system.query_log`, `system.text_log`, `system.processes` and `system.clusters` become denied. qan-api2 is unaffected, as its user holds implicit full privileges.

Raising the PMM-side change on percona/pmm#5747.

theTibi (approved):

4nte (approved):

charts/pmm-ha/templates/clickhouse-cluster.yaml
ademidoff: PMM-side change is in: percona/pmm@33e77f462 on percona/pmm#5747 flips `select_from_system_db_requires_grant` to `true` in `default-config.xml`, `low-memory-config.xml` and `dev/clickhouse-config.xml`, with the reasoning in a comment above the setting.

Verified beforehand on stock `clickhouse/clickhouse-server:25.3.6.56`, driving PMM's real `default-users.xml` so the profile, constraint and grant are the shipped ones rather than an approximation:

| Check (as the data source user) | flag `false` | flag `true` |
|---|---|---|
| `system.query_log` | readable | `Code: 497` |
| `system.text_log` | readable | `Code: 497` |
| `system.processes` | readable | `Code: 497` |
| `system.clusters` | readable | `Code: 497` |
| `system.tables` / `columns` / `databases` / `one` | ok | ok |
| `SHOW DATABASES`, `DESCRIBE TABLE`, `version()` | ok | ok |
| `SELECT ... SETTINGS max_execution_time = 30` | ok | ok |
| `system.clusters` as the privileged qan-api2 user | ok | ok |

`select_from_information_schema_requires_grant` stays off, since `information_schema` exposes metadata views only and the plugin may use them. Thanks for catching this — it was the one part of the invariant that did not hold.

maxkondr (approved):

PR #926, opened 2026-08-11
K8SPG-1115: add `spec.proxy.pgbouncer.paused`

mayankshah1607:


charts/pg-db/templates/cluster.yaml
```diff
@@ -380,6 +380,7 @@ spec:
       image: {{ .Values.proxy.pgBouncer.image }}
       replicas: {{ .Values.proxy.pgBouncer.replicas }}
       exposeSuperusers: {{ .Values.proxy.pgBouncer.exposeSuperusers | default false }}
+      pause: {{ .Values.proxy.pgBouncer.pause | default false }}
```
DhruthiKV: IIUC, this should be `paused: {{ .Values.proxy.pgBouncer.paused | default false }}` to match `spec.proxy.pgBouncer.paused` in the CRD. We should also update values.yaml and the README accordingly.

DhruthiKV (changes requested):

charts/pg-db/templates/cluster.yaml
mayankshah1607: Yep, my bad. https://github.com/percona/percona-helm-charts/pull/926/commits/e9fe5ceec97673f14ca1a8678ef807b3ec422a49

DhruthiKV (approved):
Verified changes locally, LGTM.

PR #933, opened 2026-08-20
PMM-14787 Data retention in HA

4nte:
[PMM-14787](https://perconadev.atlassian.net/browse/PMM-14787)

Related PR: https://github.com/percona/pmm/pull/5792
Feature build: https://github.com/Percona-Lab/pmm-submodules/pull/4532

`dataRetentionDays` becomes the only way to set retention, defaulting to 30 days for metrics and Query Analytics.

[PMM-14787]: https://perconadev.atlassian.net/browse/PMM-14787?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/pmm-ha/templates/_helpers.tpl
+++ b/charts/pmm-ha/templates/_helpers.tpl
@@ -292,3 +292,51 @@ dict with "name" and "value".
 {{- fail (printf "%s must match ^[A-Za-z_][A-Za-z0-9_-]*$ to be usable in the ClickHouse users.d drop-in, got %q" .name .value) -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Data retention.
+
+PMM owns retention. It applies whatever is set in the UI to Query Analytics directly, and
+to metrics by updating VMCluster.spec.retentionPeriod, which is why this chart neither
+declares that field nor sets PMM_DATA_RETENTION by default. Declaring either one would
+make every `helm upgrade` revert what the user set in the UI.
+
+`dataRetentionDays` is the opt-out: setting it pins retention declaratively by rendering
+PMM_DATA_RETENTION, which PMM treats as authoritative and which makes the UI field
+read-only. PMM requires that value to be a whole multiple of 24h and refuses to start
+otherwise, which is why this chart takes whole days and does the conversion itself rather
+than accepting a free-form duration.
+
+Example: dataRetentionDays: 90 -> PMM_DATA_RETENTION "2160h", and PMM sets the VMCluster
+retentionPeriod to "90d".
+*/}}
+{{- define "pmm.dataRetention.validate" -}}
+{{- with .Values.victoriaMetrics }}
+  {{- with .vmstorage }}
+    {{- if hasKey . "retentionPeriod" }}
+      {{- fail "victoriaMetrics.vmstorage.retentionPeriod is no longer used: PMM sets the VMCluster retention period from its own data retention setting, so a value declared here would be reverted on every `helm upgrade`. Change data retention in the PMM UI, or pin it with the top-level `dataRetentionDays` (whole days)." }}
+    {{- end }}
+  {{- end }}
+{{- end }}
+{{- with .Values.pmmEnv }}
+  {{- if hasKey . "PMM_DATA_RETENTION" }}
+    {{- fail "pmmEnv.PMM_DATA_RETENTION is managed by this chart. Set the top-level `dataRetentionDays` (whole days) instead, or leave it unset to manage retention from the PMM UI." }}
+  {{- end }}
+{{- end }}
+{{- if not (kindIs "invalid" .Values.dataRetentionDays) }}
+  {{- if or (lt (int .Values.dataRetentionDays) 1) (ne (float64 .Values.dataRetentionDays) (float64 (int .Values.dataRetentionDays))) }}
+    {{- fail (printf "dataRetentionDays must be a whole number of days greater than or equal to 1, got %v" .Values.dataRetentionDays) }}
+  {{- end }}
+{{- end }}
+{{- end -}}
+
+{{/*
+Retention period for the PMM_DATA_RETENTION environment variable, as a Go duration.
+Empty when retention is not pinned, in which case the caller omits the variable.
+*/}}
+{{- define "pmm.dataRetention.pmm" -}}
+{{- include "pmm.dataRetention.validate" . -}}
+{{- if not (kindIs "invalid" .Values.dataRetentionDays) -}}
+{{- printf "%dh" (mul (int .Values.dataRetentionDays) 24) -}}
+{{- end -}}
+{{- end -}}
--- a/charts/pmm-ha/templates/_helpers.tpl
+++ b/charts/pmm-ha/templates/_helpers.tpl
@@ -297,9 +297,9 @@ dict with "name" and "value".
 Data retention.
 
 PMM owns retention. It applies whatever is set in the UI to Query Analytics directly, and
-to metrics by updating VMCluster.spec.retentionPeriod, which is why this chart neither
-declares that field nor sets PMM_DATA_RETENTION by default. Declaring either one would
-make every `helm upgrade` revert what the user set in the UI.
+to metrics by updating VMCluster.spec.retentionPeriod, which is why this chart leaves that
+field undeclared unless dataRetentionDays pins it. Declaring it otherwise would have each
+`helm upgrade` and PMM overwrite each other, rolling vmstorage each round.
 
 `dataRetentionDays` is the opt-out: setting it pins retention declaratively by rendering
 PMM_DATA_RETENTION, which PMM treats as authoritative and which makes the UI field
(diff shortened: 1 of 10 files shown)


charts/pmm-ha/templates/vmcluster.yaml
```diff
@@ -25,12 +25,11 @@ spec:
...
-  # Retention period for metrics
-  {{- /* dataRetentionDays is opt-in and wins when set, aligning metrics with QAN. Left
-         empty - the default - metrics keep whatever vmstorage.retentionPeriod says, so an
-         upgrade never silently shortens retention and deletes data. */}}
-  retentionPeriod: {{ if .Values.dataRetentionDays }}{{ printf "%dd" (int .Values.dataRetentionDays) | quote }}{{ else }}{{ .Values.victoriaMetrics.vmstorage.retentionPeriod | default "90d" | quote }}{{ end }}
-  
+  # Metrics retention, rendered from the same dataRetentionDays as PMM_DATA_RETENTION on the
+  # PMM replicas so that metrics and Query Analytics keep data for the same period. Always
+  # declared, so an upgrade never drops the field and leaves vmstorage on the operator default.
+  retentionPeriod: {{ include "pmm.dataRetention.vm" . | quote }}
```
JiriCtvrtka, line 31: HIGH: default metrics retention drops 90d → 30d. Base renders retentionPeriod: "90d", PR renders "30d"; a plain helm upgrade on stock values makes vmstorage delete 60 days of metrics, irreversibly and with no note in README/SIZING. This is exactly what the deleted guard comment ("an upgrade never silently shortens retention and deletes data") existed to prevent.

charts/pmm-ha/templates/statefulset.yaml
```diff
@@ -284,14 +284,13 @@ spec:
...
-            {{- if .Values.dataRetentionDays }}
+            # Derived from dataRetentionDays, and rendered on every replica unconditionally:
+            # PMM fixes retention when it starts, so a replica without this variable falls
+            # back to whatever period is already stored and warns that the chart should be
+            # supplying one. Declared here rather than in pmmEnv so it always wins over the
+            # ConfigMap.
             - name: PMM_DATA_RETENTION
-              value: "{{ mul (int .Values.dataRetentionDays) 24 }}h"
-            {{- end }}
+              value: {{ include "pmm.dataRetention.pmm" . | quote }}
```
JiriCtvrtka, line 300: HIGH: PMM_DATA_RETENTION is now emitted unconditionally at the chart default. Base emitted it never; PR emits "720h". An install whose admin had raised QAN retention in the PMM UI (the documented path on the old chart) has it forced back to 30d on upgrade, pruning ClickHouse. Same failure mode the PR's own commit 1da3674 describes, with its mitigation dropped in the final state.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -508,3 +508,62 @@ and be rejected outright.
...
+      {{- fail "victoriaMetrics.vmstorage.retentionPeriod is no longer used: this chart renders the VMCluster retention period from the top-level `dataRetentionDays`, and a value declared here would let metrics keep data for a different period than Query Analytics. Set `dataRetentionDays` (whole days) instead." }}
+    {{- end }}
+  {{- end }}
+{{- end }}
+{{- with .Values.pmmEnv }}
+  {{- if hasKey . "PMM_DATA_RETENTION" }}
+    {{- fail "pmmEnv.PMM_DATA_RETENTION is derived by this chart from the top-level `dataRetentionDays`, and a value declared here would let Query Analytics keep data for a different period than metrics. Set `dataRetentionDays` (whole days) instead." }}
+  {{- end }}
+{{- end }}
+{{- if kindIs "invalid" .Values.dataRetentionDays }}
```
JiriCtvrtka: MEDIUM: kindIs "invalid" only catches nil, so the previous chart's shipped default dataRetentionDays: "" skips the good "must be set" message and fails with the truncated ...greater than or equal to 1, got (verified). The new CI matrix covers null/0/1.5/true but not the empty string the chart itself used to ship.

charts/pmm-ha/templates/statefulset.yaml
4nte: We've decided to remove the data retention setting from the PMM UI, so it can only be set through PMM_DATA_RETENTION. The chart therefore always renders it from dataRetentionDays. Leaving it unset would leave each replica on whatever value it had stored, with no way to change it. The upgrade case you describe (a value raised in the old UI) is covered by an upgrade note in the README telling users to carry that value into dataRetentionDays first.

charts/pmm-ha/templates/vmcluster.yaml
4nte: This was intentional, standalone PMM uses a single retention for metrics and QAN, defaulting to 30 days (DataRetention in pmm-managed feeds both VictoriaMetrics' --retentionPeriod and qan-api2). The Technical Preview HA chart was the outlier: 90d for metrics via its own vmstorage default, while QAN was already on PMM's 30d. dataRetentionDays: 30 makes HA match standalone. However the upgrade concern is real, so the README now has an upgrade note for Technical Preview installations: "check the current period and set dataRetentionDays: 90 before upgrading to keep it"

charts/pmm-ha/templates/_helpers.tpl
4nte: Fixed in 35a8ac4b. An empty or whitespace-only `dataRetentionDays` is now treated as unset, so it gets the "must be set" message instead of falling through to the range check. I added the case to `tests/values-validation_test.yaml`, which is where the rejection matrix lives now that it has moved from the workflow into helm unittest.

JiriCtvrtka (approved):
Just two non blocking things:
1) _helpers.tpl:683,689 — hasKey rejects --set victoriaMetrics.vmstorage.retentionPeriod=null. The README's actual instruction ("remove it from your values") works correctly; only the explicit-null --set form aborts. Worth a follow-up, doesn't block.
2) _helpers.tpl:700 — days * 24 int64 overflow needs an input like 400000000000000000 to trigger. Cosmetic robustness; an upper bound (say ≤ 3650) would close it cheaply if you're touching the file for the version bump anyway.

4nte:
Thanks @JiriCtvrtka, both addressed in 94b98474:

1. An explicit null on `victoriaMetrics.vmstorage.retentionPeriod` or `pmmEnv.PMM_DATA_RETENTION` now counts as removed, not declared. Adding the test showed a second issue: a null `pmmEnv` entry used to reach the ConfigMap as `PMM_DATA_RETENTION: null`. `configmap.yaml` now drops null entries. The default render is byte-identical to before.
2. `dataRetentionDays` is capped at 36500. I went with 100 years instead of 3650 because that is VictoriaMetrics' own maximum retention, so the cap closes the overflow without adding a product limit.

Both have helm unittest cases (56/56 passing).

ademidoff (approved):
The PMM docs page linked from the new "Data retention" section in charts/pmm-ha/README.md (L767) contradicts this chart.

[Adjust data retention and other settings](https://docs.percona.com/percona-monitoring-and-management/3/install-pmm/install-HA-clustered.html#adjust-data-retention-and-other-settings), sourced from documentation/docs/install-pmm/install-HA-clustered.md (L718–734) in percona/pmm, currently:

tells users to set pmmEnv.PMM_DATA_RETENTION: "2160h", which this chart now rejects at render time
gives the default as 90 days, where the chart now defaults to 30
doesn't cover what the README sends readers there for: how to change retention, how to check the period a replica is using, or what shortening it does to existing data

[percona/pmm#5792](https://github.com/percona/pmm/issues/5792) doesn't touch the documentation, so as things stand anyone following the published docs will hit the new pmmEnv.PMM_DATA_RETENTION failure. Could we get a docs PR updating that section to dataRetentionDays and land it together with this one? The Technical Preview known-issue row for PMM-14709 on the same page (L1105) may also need revisiting.

4nte:
@ademidoff Thanks, good catch. The docs PR for that page is percona/pmm#5963 (against `doc-3.10.0`, since user docs ship ahead of the code), and it lands together with this one and percona/pmm#5792. In [Adjust data retention and other settings](https://github.com/percona/pmm/pull/5963/files) it:

- replaces the `pmmEnv.PMM_DATA_RETENTION: "2160h"` instructions with the top-level `dataRetentionDays` and corrects the default to 30 days
- covers how to change retention (`dataRetentionDays` + `helm upgrade`), how to check the period each replica is using (the `Data retention:` line in `pmm-managed.log`), and what shortening it does to existing data
- adds a `dataRetentionDays` row to the Helm parameter reference
- drops the PMM-14709 Technical Preview known-issue row, and the PMM-14787 known issue on the HA overview page

I also just added (f9dbc7d7d) a Technical Preview upgrade warning there, matching the one in this README, since the 90d to 30d default change is the part most likely to catch people out.

PR #935, opened 2026-08-24
PMM-14403 Finalize the HA resource limitations

theTibi:
Sizes the `pmm-ha` chart's defaults for **~100 monitored nodes at 30-day retention** and gives every container explicit requests and limits.

## Resources

Several components had no resources at all and ran as BestEffort — the PostgreSQL instances, pgBouncer, the PMM client sidecar, kube-state-metrics and node-exporter. That makes the pods holding PMM's inventory, Grafana state and alert rules the first thing the kubelet evicts under node pressure.

HAProxy carries every client `remote_write`, QAN payload and UI request, and terminates TLS for all of them. It inherited a request but **no limit at all** from its subchart.

| Component | Now |
|---|---|
| PMM server | req `2`/`4Gi` · lim `4`/`8Gi` |
| HAProxy | req `250m`/`128Mi` · lim `1`/`512Mi` |
| ClickHouse | req `1`/`4Gi` · lim `4`/`8Gi` |
| CH Keeper | req `250m`/`512Mi` · lim `1`/`1Gi` |
| vmselect | req `500m`/`1Gi` · lim `2`/`4Gi` |
| vmstorage | req `500m`/`2Gi` · lim `2`/`4Gi` |
| vmagent | req `250m`/`512Mi` · lim `1`/`1Gi` |
| PostgreSQL | req `500m`/`1Gi` · lim `2`/`4Gi` |
| pgBouncer | req `100m`/`128Mi` · lim `500m`/`512Mi` |
| PG PMM sidecar | req `100m`/`128Mi` · lim `500m`/`512Mi` |
| kube-state-metrics | req `100m`/`128Mi` · lim `500m`/`512Mi` |
| node-exporter | req `50m`/`64Mi` · lim `200m`/`128Mi` |

vminsert and vmauth were already correct and are unchanged.

The wide request/limit gaps in the write path are deliberate. Only the HAProxy-elected leader serves traffic, and every PMM client buffers up to 1 GB on disk during an outage and replays it on reconnect — a failover or rolling upgrade releases that burst all at once.

## `dataRetentionDays` (optional)

Metrics and Query Analytics are retained by two independent mechanisms, and in HA neither knows about the other:

- **Metrics** — `victoriaMetrics.vmstorage.retentionPeriod`, enforced by the VictoriaMetrics operator. Default `90d`.
- **Query Analytics** — PMM's own setting, enforced by `qan-api2` against ClickHouse, changeable from the PMM UI. Default 30 days.

`PMM_DATA_RETENTION` governs only the second one here: the built-in VictoriaMetrics that would otherwise consume it doesn't run in an HA deployment. Setting it also *locks* the setting — `ChangeSettings` returns `FailedPrecondition` while the variable is present.

`dataRetentionDays` is therefore **empty by default**, and the chart touches neither mechanism. Set it to drive both from one value, which is usually what a declaratively managed deployment wants.

| Config | vmstorage | `PMM_DATA_RETENTION` |
|---|---|---|
| default (empty) | `90d` | not set |
| `dataRetentionDays=90` | `90d` | `2160h` |
| `dataRetentionDays=30` | `30d` | `720h` |
| `retentionPeriod=180d` only | `180d` | not set |
| both set (45 / 180d) | `45d` | `1080h` |
| `retentionPeriod=""` only | `90d` | not set |

When set, `dataRetentionDays` takes precedence over `vmstorage.retentionPeriod`.

## Documentation

- `docs/SIZING.md` — the planning model, series-per-node by database type, the storage formulas, per-component tables for 100/500/1000 nodes, and the PromQL/SQL to re-measure the constants against your own workload. The constants are measured from a running cluster, not estimated.
- `examples/values-500-nodes.yaml`, `examples/values-1000-nodes.yaml` — ready profiles for larger fleets.
- README gains a Sizing section and two prerequisites: a storage class with `allowVolumeExpansion: true`, and `amd64` nodes for the PMM and ClickHouse pods (QAN needs SSE4.2 and PMM Server has no native ARM64 build).

## Upgrade notes

- **ClickHouse PVC moves from 20Gi to 50Gi.** Each replica holds a full copy of Query Analytics, and 20Gi does not cover 100 services at 30-day retention once ClickHouse's own `system.*` tables are accounted for. This is a PVC expansion on upgrade and requires `allowVolumeExpansion: true`.
- **No retention change.** `dataRetentionDays` is empty by default and metrics stay on `90d`, so upgrades do not shorten retention or delete data. Query Analytics stays on PMM's own setting and remains adjustable in the UI.
- **Pre-existing, documented but not changed:** the default 50Gi vmstorage volume holds roughly 30 days at 100 nodes, not the 90 that the default retention asks for. `docs/SIZING.md` explains the two ways to resolve it — neither default is changed here, since one deletes data and the other forces a PVC expansion on every install.
- No other PVC default changes — sizes are only ever raised, never lowered, since PVCs cannot shrink.

## Verification

`helm lint` and `helm template` pass against the default values, both example profiles and `ci/ci-values.yaml`. Rendered output confirmed for every container having both requests and limits, correct PVC sizes, retention wiring across all six configurations above, and both example profiles applying end to end.


charts/pmm-ha/values.yaml
```diff
@@ -883,16 +981,25 @@ victoriaMetrics:
     ## @param victoriaMetrics.vmstorage.replicaCount Number of vmstorage replicas
     replicaCount: 3
     ## @param victoriaMetrics.vmstorage.retentionPeriod Data retention period
-    retentionPeriod: "90d"
+    ## Leave empty to inherit the chart-wide dataRetentionDays, which keeps metrics
+    ## and Query Analytics on the same retention. Set explicitly (e.g. "90d") only
+    ## when metrics deliberately need to outlive QAN.
+    retentionPeriod: ""
```
JiriCtvrtka: I’m not very familiar with HA, but I think this is a real-life scenario.

Empty resolves to the operator default 30d, so every install that took the default silently drops from 90 to 30 days of metrics on the next helm upgrade. Deleted TSDB data doesn't come back, and the PR description asserts the opposite (only explicit pinners affected). Either keep "90d" or derive it from dataRetentionDays.

theTibi: good point but at this point no one should use this in production because it is not GA yet. :) But I can make it 90d

JiriCtvrtka (approved):

charts/pmm-ha/values.yaml
JiriCtvrtka: I see, then it is fine. About 90d default it is up to you.

theTibi: You're right, and thanks for catching it — fixed in 946e4af.

Small correction on the mechanism: empty didn't fall through to an operator default, the chart substituted `dataRetentionDays` (30) before the operator saw it. Same outcome though, and my commit message did say "installs already pinning it keep their behaviour", which reads as though everyone else was fine. That was wrong.

Digging into it turned up two things that make the single-value design worse than it looked:

**`PMM_DATA_RETENTION` only governs Query Analytics in HA.** The built-in VictoriaMetrics that would otherwise consume it doesn't run in an HA deployment — `supervisorctl status` on a PMM pod shows only `qan-api2`, which prunes ClickHouse. Metrics retention belongs entirely to the operator, so the two were never really one knob.

**Setting `PMM_DATA_RETENTION` locks the setting.** `ChangeSettings` returns `FailedPrecondition` for any retention change while the variable is present ([server.go#L486-L488](https://github.com/percona/pmm/blob/main/managed/services/server/server.go#L486-L488)). The chart was quietly removing a control users have in the PMM UI today.

So flipping the default to 90 wouldn't have been right either — it stops the metrics loss but moves QAN from 30d to 90d, tripling ClickHouse storage against a 50Gi PVC sized for 30d, and still locks the UI.

What it does now: `dataRetentionDays` is **empty by default**. Left empty the chart touches neither mechanism — `vmstorage.retentionPeriod` is back to a literal `"90d"` and `PMM_DATA_RETENTION` isn't emitted, so QAN stays on PMM's own setting and remains adjustable in the UI. Upgrade is a no-op. Set it and it drives both, taking precedence over `vmstorage.retentionPeriod`.

Verified across six configurations:

| Config | vmstorage | `PMM_DATA_RETENTION` |
|---|---|---|
| defaults | `90d` | not set |
| `dataRetentionDays=90` | `90d` | `2160h` |
| `dataRetentionDays=30` | `30d` | `720h` |
| `retentionPeriod=180d` only | `180d` | not set |
| both set (45 / 180d) | `45d` | `1080h` |
| `retentionPeriod=""` only | `90d` | not set |

One related thing worth flagging, now documented in `SIZING.md` rather than changed: the default 50Gi vmstorage volume holds roughly 30 days at 100 nodes, not the 90 the default retention asks for. That mismatch predates this PR. I left both defaults alone because lowering retention deletes data and raising the PVC forces an expansion on every install — the doc spells out which knob to turn instead.

PR #936, opened 2026-08-27
PMM-14738 Validate PMM HA replica counts and publish the supported range

theTibi:
[PMM-14738](https://perconadev.atlassian.net/browse/PMM-14738)

Adds the validation and documentation for PMM HA scaling that the chart never had, plus the test suite that guards it.

## Nothing checked the two values that can break a cluster

Rendered at `replicas` 1 through 12, everything was accepted silently:

| `replicas` | Raft peers | HAProxy slots | Warning |
|---|---|---|---|
| 2 | 2 | 10 | none |
| 4 | 4 | 10 | none |
| 12 | **12** | **10** | none |

* **Even counts.** Raft elects by majority, so 4 replicas need more votes to elect a leader without surviving more losses than 3 do, and `replicas=2` survives none — one pod restart stops the cluster.
* **Above `maxReplicas`.** HAProxy renders `server-template pmm 1-<maxReplicas>` and fills the slots from a headless-service DNS answer in arbitrary order, then marks a backend UP only when it answers `/v1/server/leaderHealthCheck`. A pod without a slot is invisible to HAProxy — and if the Raft leader lands there, **every backend is DOWN and PMM serves 503**. Not a degraded pod: an outage.

`pmm.replicas.validate` and `pmm.keeper.validate` now reject both, plus malformed input. The regex is strict on purpose: sprig's `int` parses base 0, so `"010"` would have validated as 8, and anything wider than int64 overflows to 0 — either way the error would quote a number nobody typed.

Suggestions are clamped to the ceiling, so the error never names a value the next check rejects:

```
replicas=2   Use 1 or 3.
replicas=4   Use 3 or 5.
replicas=10  Use 9.
replicas=12  maxReplicas is 10, so the largest supported value is 9.
```

`clickhouse.keeper.replicasCount` is a Raft ensemble too. `values.yaml` has always said it "should be odd"; that is now enforceable rather than advisory.

## Smaller fixes

* `maxReplicas` had its `| default 10` duplicated between the validator and `haproxy-configmap.yaml`. Both now read `pmm.maxReplicas`, so they cannot drift — and because `default` treats 0 as empty, `maxReplicas: 0` used to become 10 silently.
* The workflow's `paths:` filter did not include the workflow file, so a PR editing only its assertions ran no job.
* `azure/setup-helm` v4.2.0 / Helm 3.15.4 -> v5.0.0 / Helm 3.21.0, which is what the `helm-unittest` plugin needs to load.

## Tests

`charts/pmm-ha/tests/` — 27 helm-unittest cases, **no cluster required**, runnable with `make test-pmm-ha`:

```
Test Suites: 3 passed, 3 total
Tests:       27 passed, 27 total
```

Scoping a suite to `templates/statefulset.yaml` skips the `lookup` in `pg-user-credentials-secrets.yaml` entirely, so these do not depend on Helm's template ordering the way a `helm template` assertion would.

Mutation-tested against the merged tree — every regression is caught:

| Mutation | Result |
|---|---|
| revert `pmm.haPeers` to `.Release.Name` | 2 failed |
| revert the `server-template` slot count to a literal `10` | 2 failed |
| drop the even-count check | 5 failed |
| make `pmm.validate.oddCount` always reject | 20 failed, 5 errored |
| drop the `maxReplicas` ceiling check | 2 failed |

## Merged with PMM-HA-GA

Merging the current base branch dropped two changes this PR originally carried, because equivalent work landed first:

* **The `PMM_HA_PEERS` fix.** `pmm.haPeers` built peer hostnames from `.Release.Name` while the StatefulSet is named after `pmm.fullname`, and the two are equal only when the release name already contains the chart name — so `helm install pmm-prod percona/pmm-ha` gave Raft peers that never resolved. #868 (PMM-15151) shipped the same fix, so there is nothing left to change here. `tests/ha-peers_test.yaml` stays as the regression guard, including a case asserting that a release named `pmm-ha` still renders byte-identically.
* **`ci/ci-values.yaml`.** This PR removed `clickhouse.keeper.replicas`, a key nothing reads (the template reads `replicasCount`). #946 (PMM-15441) rewrote the file and already sets `replicasCount`.

`templates/haproxy-init-configmap.yaml` was deleted by #937 (PMM-15394), and this PR's `pmm.fullname` fix to its host loop went with it. The chart version conflict resolved to `1.7.0`, keeping `appVersion "3.9.1"` from #937.

## Upgrade impact

**Breaking for anyone running an even `replicas`.** `2` or `4` was accepted before and now fails the render; those users must set an odd value in the same `helm upgrade`, which recreates the pods. The README carries an upgrade note.

Chart `1.6.2` -> `1.7.0`. **Reviewers: a case can be made for `2.0.0`** — a `^1.6.2` consumer will otherwise adopt a render-breaking change automatically. Happy to bump if you prefer.

## Related

* Documentation half: percona/pmm#5842

[PMM-14738]: https://perconadev.atlassian.net/browse/PMM-14738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charts/pmm-ha/README.md
```diff
@@ -862,6 +888,53 @@ Common troubleshooting steps for PMM HA:
...
+- **Raising `maxReplicas` needs an HAProxy restart.** It is rendered into the
+  `pmm-ha-haproxy` ConfigMap, which the chart does not roll on upgrade. Routing itself
+  is DNS-based, so changing `replicas` needs no restart. Bump
+  `haproxy.podAnnotations."pmm.percona.com/config-version"` in the same `helm upgrade`
+  so the pods restart and pick up the new `server-template`:
+
+  ```sh
+  helm upgrade pmm-ha percona/pmm-ha --namespace pmm \
+    --set maxReplicas=20 \
+    --set-string 'haproxy.podAnnotations.pmm\.percona\.com/config-version=4'
```
ademidoff: This example is a no-op. `4` is already the default in `charts/pmm-ha/values.yaml:660`, so running this verbatim changes nothing about the pod template — HAProxy never restarts and keeps serving the old `server-template` slot count, which is precisely the failure the bullet above tells the reader to avoid.

Wants a different value, e.g. `5`:

```suggestion
    --set-string 'haproxy.podAnnotations.pmm\.percona\.com/config-version=5'
```

For what it's worth I did verify the underlying claim before flagging this: `haproxy.configmap.enabled: false` plus `haproxy.config: ""` means the ConfigMap comes from this chart while the subchart's own checksum annotation is computed from the (empty) `config` value, so nothing rolls the pods when `maxReplicas` changes. The advice is right — only the examples don't carry it out.

charts/pmm-ha/README.md
```diff
@@ -862,6 +888,53 @@ Common troubleshooting steps for PMM HA:
...
+  ```
+
+  Or in `values.yaml`:
+
+  ```yaml
+  maxReplicas: 20
+  haproxy:
+    podAnnotations:
+      pmm.percona.com/config-version: "4"
```
ademidoff: Same no-op as the `--set-string` example above — this matches the default in `charts/pmm-ha/values.yaml:660`, so it wouldn't restart HAProxy either.

```suggestion
      pmm.percona.com/config-version: "5"
```

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -303,6 +303,103 @@ Called from statefulset.yaml, which always renders.
...
+{{- $lower := sub $n 1 -}}
+{{- $upper := add $n 1 -}}
+{{- $hint := printf "Use %d or %d." $lower $upper -}}
+{{- if gt $ceiling 0 -}}
+{{- $maxOdd := $ceiling -}}
+{{- if eq (mod $ceiling 2) 0 -}}
+{{- $maxOdd = sub $ceiling 1 -}}
+{{- end -}}
+{{- if le $upper $maxOdd -}}
+{{- $hint = printf "Use %d or %d." $lower $upper -}}
```
ademidoff: Dead assignment: `$hint` already holds this exact string from line 347, so this branch is a no-op.

Dropping it leaves the clamp reading as what it actually does — keep the default hint, narrow to the lower neighbour, or report the ceiling:

```
{{- if gt $ceiling 0 -}}
{{- $maxOdd := $ceiling -}}
{{- if eq (mod $ceiling 2) 0 -}}
{{- $maxOdd = sub $ceiling 1 -}}
{{- end -}}
{{- if gt $upper $maxOdd -}}
{{- if le $lower $maxOdd -}}
{{- $hint = printf "Use %d." $lower -}}
{{- else -}}
{{- $hint = printf "%s is %d, so the largest supported value is %d." (.ceilingName | default "The ceiling") $ceiling $maxOdd -}}
{{- end -}}
{{- end -}}
{{- end -}}
```

Purely cosmetic — the behaviour is correct as written, I checked all three branches.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -303,6 +303,103 @@ Called from statefulset.yaml, which always renders.
...
+Fail-fast validation for the ClickHouse Keeper node count.
+Called from statefulset.yaml alongside the other value checks, for the same ordering
+reason described above.
+
+The parenthesised lookup matches pmm.nodeExporter.mode: without it, a nulled clickhouse
+or clickhouse.keeper key aborts with a raw Go nil-pointer error instead of the message
+this validator exists to produce.
+*/}}
+{{- define "pmm.keeper.validate" -}}
+{{- include "pmm.validate.oddCount" (dict "name" "clickhouse.keeper.replicasCount" "value" ((.Values.clickhouse).keeper).replicasCount) -}}
```
ademidoff: Keeper is left unbounded here — no `ceiling` passed, so `pmm.validate.oddCount` only enforces parity and the 1-9999 regex. Verified: `clickhouse.keeper.replicasCount: 9999` renders happily and the operator would be asked for 9999 Keeper pods.

That's consistent with there being no HAProxy-style hard constraint on Keeper, so it's a judgement call rather than a bug. But a Keeper ensemble past 5-7 is pure cost, and `9999` is the kind of typo that eats a cluster before anyone notices. Passing a small ceiling reuses the clamp logic you already wrote and gets a good message for free:

```
{{- include "pmm.validate.oddCount" (dict "name" "clickhouse.keeper.replicasCount" "value" ((.Values.clickhouse).keeper).replicasCount "ceiling" 9 "ceilingName" "The supported maximum") -}}
```

ademidoff (commented):
Ran the suites and a real `helm template` against `6c35606`: 27/27 pass, and the validators do fire before the `lookup` in `charts/pmm-ha/templates/pg-user-credentials-secrets.yaml`, so the error a user sees is the actionable one rather than a red herring about a missing secret. I also probed the values the suite doesn't cover — `1.5`, `true`, `null`, `""`, `8/8`, `2/1`, and the `replicas == maxReplicas` boundary — and all behave as designed.

Start with the two `config-version` examples in `charts/pmm-ha/README.md`; they're the one thing here that misleads rather than merely reads oddly.

charts/pmm-ha/values.yaml
```diff
@@ -314,12 +314,26 @@ service:
...
-## maximum number of server slots HAProxy will create. Adjust if you plan to
-## scale beyond 10 replicas.
+## maximum number of server slots HAProxy will create, and therefore the ceiling on
+## replicas - the chart rejects a replicas above it.
+##
+## Raising this alone is not enough: it is rendered into the HAProxy ConfigMap, which
+## the chart does not roll on upgrade. Bump haproxy.podAnnotations
+## "pmm.percona.com/config-version" in the same upgrade so HAProxy restarts and picks
+## up the new server-template.
 maxReplicas: 10
```
ademidoff, line 337: Minor: `10` is even, so under the new odd-only rule slot 10 is permanently unreachable and the real ceiling is 9.

The validator handles it correctly — `replicas=12` gives "maxReplicas is 10, so the largest supported value is 9", which is exactly right — but an odd default would make the documented ceiling and the effective ceiling agree without the reader having to work it out. `11` keeps 9 reachable and adds a little headroom; `9` makes the two numbers identical.

Not worth churning an existing value over if you'd rather leave it; the asymmetry is just slightly surprising now that it's enforced.

.github/workflows/pmm-ha-pr-checks.yaml
```diff
@@ -4,6 +4,8 @@ on:
     paths:
       - 'charts/pmm-ha/**'
       - 'charts/pmm-ha-dependencies/**'
+      # Without this a PR that only edits the assertions below runs no job at all.
+      - '.github/workflows/pmm-ha-pr-checks.yaml'
```
ademidoff, line 8: Good catch on the workflow file itself — `Makefile` has exactly the same hole. `make test-pmm-ha` now gates this chart, but a `Makefile`-only PR matches none of these paths, and `.github/workflows/unit-test.yaml` (which *does* filter on `Makefile`) only runs `test-pxc-operator` / `test-pxc-db` in its matrix. So breaking the `test-pmm-ha` target slips through both workflows.

```suggestion
      - '.github/workflows/pmm-ha-pr-checks.yaml'
      # Same reason: `make test-pmm-ha` gates this chart, and unit-test.yaml's matrix
      # does not cover it.
      - 'Makefile'
```

charts/pmm-ha/tests/values-validation_test.yaml
```diff
@@ -0,0 +1,128 @@
+# Rejection cases for the value validators wired into statefulset.yaml.
+#
+# The suite renders statefulset.yaml alone on purpose. Every case is expected to abort in
+# a validator, so nothing downstream matters - and keeping pg-user-credentials-secrets.yaml
+# out means these run with no cluster and no dependency on Helm's template ordering.
```
ademidoff, line 5: Worth spelling out one more consequence of the scoping, because it's a trap: this suite works with `templates/statefulset.yaml` alone *only* because every case aborts in a validator on lines 1-3. The moment someone adds a `notFailedTemplate` case here it errors instead of passing:

```
error calling include: template: no template "pmm-ha/templates/configmap.yaml"
associated with template "gotpl"
```

— because `charts/pmm-ha/templates/statefulset.yaml:21` includes `pmm.podAnnotations`, which reaches for `configmap.yaml`. (`charts/pmm-ha/tests/ha-peers_test.yaml` lists the extra templates, which is why its two positive cases work.) A sentence here saves the next person that debugging session.

Two boundary cases I'd also pin while you're in here — I checked both by hand and they're correct, they're just unpinned:

- `replicas: 3, maxReplicas: 3` renders. The ceiling check is `gt`, not `ge`, and nothing currently asserts that `replicas == maxReplicas` is allowed.
- `maxReplicas: 101` is rejected. Only `200` is covered, and `^([1-9][0-9]?|100)$` is exactly the sort of regex someone later "simplifies" to `[0-9]{1,3}` — at which point `101` starts rendering a 101-slot `server-template` and only the off-by-one case would have caught it.

charts/pmm-ha/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: pmm-ha
 description: A Helm chart for Percona Monitoring and Management (PMM)
 type: application
-version: 1.6.2
+version: 1.7.0
```
ademidoff, line 5: Since `pmm-ha` is still tech preview and the version bump is planned for GA, I don't think this line belongs in this PR.

`CLAUDE.md` spells out the convention for the PMM-owned charts: don't bump `version:` in a feature PR — changes accumulate on the branch and the bump happens once in a dedicated release PR. (That's the opposite of the `pxc-*` / `psmdb-*` charts, where `ct lint --check-version-increment` forces a bump per change, which is probably where the instinct comes from.)

Reverting to `1.6.2` also removes the thing that just cost you a merge conflict, and would cost the next PR into `PMM-HA-GA` the same one:

```suggestion
version: 1.6.2
```

If you drop the bump, `charts/pmm-ha/README.md:820` needs the version out of its heading too — `> **Upgrade note**` on its own carries the same information, since the note is about the behaviour change rather than a specific release.

Worth saying for the GA discussion rather than for this PR: whenever that bump does happen, this change argues for a **major**. `helm search repo percona/pmm-ha --versions` lists 1.3.23 through 1.6.1 as published, so `^1.6.x` consumers exist and an even `replicas` that used to install now fails the render. Tech preview is a fair reason not to treat that as binding, but it's the kind of thing worth deciding deliberately at GA rather than inheriting.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -303,6 +303,103 @@ Called from statefulset.yaml, which always renders.
...
+{{- end -}}
+{{- if le $upper $maxOdd -}}
+{{- $hint = printf "Use %d or %d." $lower $upper -}}
+{{- else if le $lower $maxOdd -}}
+{{- $hint = printf "Use %d." $lower -}}
+{{- else -}}
+{{- $hint = printf "%s is %d, so the largest supported value is %d." (.ceilingName | default "The ceiling") $ceiling $maxOdd -}}
+{{- end -}}
+{{- end -}}
+{{- fail (printf "%s must be odd so Raft can form a quorum, got %d: an even count needs more votes to elect a leader without surviving more failures. %s" $name $n $hint) -}}
```
ademidoff: Two things about this message, moving them out of my review body and onto the line they concern.

**The wording overstates the problem.** "must be odd so Raft can form a quorum" isn't what's happening — an even ensemble forms quorum fine. The explanation you've already written in the same string is the accurate one, so the opening clause just contradicts it. Something like:

```suggestion
{{- fail (printf "%s must be odd, got %d: an even count adds a voter without adding fault tolerance - it needs more votes to elect a leader while surviving no more failures. %s" $name $n $hint) -}}
```

**And the policy is stricter than the rationale.** Worth being deliberate about rather than inheriting it: `2` is genuinely broken (quorum 2, tolerates zero losses, so one pod restart stops the cluster), but `4` is *suboptimal*, not broken — quorum 3, tolerates one loss, exactly like `3`. So this turns a working cluster into a failed `helm upgrade` that needs a full pod recreation to escape.

I'd still keep the hard fail: odd-only as "the supported set" is defensible, the hint is genuinely helpful, and tech preview is the right time to tighten it. Just flagging that `4` is the case where someone will push back, and it's the strongest argument for a major bump whenever the GA version lands.

charts/pmm-ha/README.md
theTibi: Fixed in 93743d1 — both examples now use `5`, and the bullet says why: the chart ships `"4"`.

charts/pmm-ha/README.md
theTibi: Fixed in 93743d1 along with the `--set-string` example above.

charts/pmm-ha/templates/_helpers.tpl
theTibi: Applied in 93743d1.

.github/workflows/pmm-ha-pr-checks.yaml
theTibi: Added in 93743d1.

charts/pmm-ha/templates/_helpers.tpl
theTibi: Bounded at 9 in 93743d1, but the one-liner alone does not do it: `ceiling` only clamps the hint printed for an even value, it never rejects. I applied your snippet verbatim and `replicasCount: 9999` still rendered.

The rejection is now explicit in `pmm.keeper.validate` — kept out of `oddCount` so it cannot fire ahead of the `maxReplicas` check in `pmm.replicas.validate` and swallow that message. `ceiling`/`ceilingName` are still passed so the even-value hint clamps. Covered by four new cases: 11 rejected, 10 → "Use 9.", 12 → ceiling message, 9 accepted.

charts/pmm-ha/templates/_helpers.tpl
theTibi: Reworded in 93743d1: "must be odd, got %d: an even count adds a Raft voter without adding fault tolerance - it widens the majority a leader election needs while surviving no more failures." Hard fail kept.

charts/pmm-ha/tests/values-validation_test.yaml
theTibi: Note added in 93743d1 with the verbatim error, and `maxReplicas: 101` is now pinned.

`replicas == maxReplicas` was already covered — `accepts every supported replica count` in `tests/ha-peers_test.yaml` sets `replicas: 9, maxReplicas: 9`, and mutating `gt` to `ge` fails it. I labelled it so it reads as the boundary it guards rather than adding a duplicate.

charts/pmm-ha/values.yaml
theTibi: Leaving `10`. Changing a published default to make the documented and effective ceilings match costs more than the asymmetry, and the validator already reports the effective one ("maxReplicas is 10, so the largest supported value is 9").

charts/pmm-ha/Chart.yaml
theTibi: Leaving `1.7.0` for now — we will settle versioning separately. Two things worth checking first: there is no `CLAUDE.md` in this repo (not on `main`, not on `PMM-HA-GA`, and code search returns nothing), and #937 bumped `1.6.1` → `1.6.2` in a feature PR into `PMM-HA-GA`. Agreed the major-at-GA question is worth deciding deliberately.

4nte (approved):

PR #938, opened 2026-08-27
PMM-15393 Let HAProxy scale beyond the worker-node count

theTibi:
## Problem

`charts/pmm-ha` set a `requiredDuringSchedulingIgnoredDuringExecution` pod anti-affinity on
`kubernetes.io/hostname` for HAProxy, so every replica had to land on a distinct node.
`haproxy.replicaCount` could therefore never exceed the worker count — replicas beyond it stayed
`Pending` forever (`didn't match pod anti-affinity rules`) while Helm still reported the upgrade as
successful.

## Change

- Replace the required anti-affinity with a `topologySpreadConstraints` entry
  (`maxSkew: 1`, `topologyKey: kubernetes.io/hostname`, `whenUnsatisfiable: ScheduleAnyway`).
  Replicas still spread one per node whenever there are enough nodes; beyond that they co-locate
  instead of failing to schedule.
- Enable a `PodDisruptionBudget` (`minAvailable: 1`) for HAProxy, so a node drain cannot take out
  every replica sharing a node. The chart previously had no PDB.
- Update the README statements that claimed anti-affinity for all components, and the scaling
  example that uses `--set haproxy.replicaCount=5`.

Both keys are supported natively by the `haproxytech/haproxy` 1.24.0 subchart, so this is a
values-only change.

PostgreSQL and pgBouncer keep their required anti-affinity: co-locating two Patroni members means a
single node loss costs quorum.

## Testing

Rendered `haproxy` with `replicaCount=5`: `replicas: 5`, the spread constraint present, no
`affinity` block, and a PDB selecting only this release. `helm lint` clean.

Scheduler A/B on EKS, both Deployments pinned to a single eligible node with 3 replicas:

| Config | Result |
| --- | --- |
| Required anti-affinity (before) | 1 Running, 2 Pending — `didn't match pod anti-affinity rules` |
| Topology spread (after) | 3/3 Running, co-located |

Unblocks the 5-replica HAProxy certification in PMM-14738.

https://perconadev.atlassian.net/browse/PMM-15393


charts/pmm-ha/values.yaml
```diff
@@ -569,14 +514,30 @@ haproxy:
...
+      whenUnsatisfiable: ScheduleAnyway
+      labelSelector:
+        matchLabels:
+          app.kubernetes.io/name: haproxy
+
+  # Keep at least one HAProxy pod serving during voluntary disruptions (node
+  # drains, upgrades), now that replicas are allowed to share a node.
+  PodDisruptionBudget:
+    enable: true
+    minAvailable: 1
```
JiriCtvrtka: **PDB makes nodes undrainable at replicaCount: 1**

`PodDisruptionBudget.enable: true` renders unconditionally, so minAvailable: 1 applies
at every replica count — including 1, where it gives allowedDisruptions = 0.
`kubectl drain` then hangs forever, cluster-autoscaler can't reclaim the node, and
managed node-pool upgrades stall. replicaCount 1 isn't hypothetical: it's the
subchart's own default (haproxy-1.24.0/values.yaml:86) and PMM-14738 explicitly
certifies "min (1) + max (5)" for haproxy.

At the default of 3 the same PDB has the opposite problem — it permits two
simultaneous evictions, so a rolling node upgrade can drop HAProxy to 1/3 while the
comment above promises to "keep at least one pod serving".

maxUnavailable: 1 expresses the intent correctly at every count — 2 of 3 stay up,
and a single replica stays drainable:

  PodDisruptionBudget:
    enable: true
    maxUnavailable: 1

4nte (approved):

PR #941, opened 2026-08-28
PMM-15401 Validate the pg-db PMM token instead of trusting the secret

theTibi:
## Problem

The token-init Job skips token creation whenever `pg-pmm-secret` holds a non-empty
`PMM_SERVER_TOKEN`, without checking that the token still authenticates — though its own
comment says "valid token".

That secret has a fixed name and is created with `kubectl` rather than by Helm, so
`helm uninstall` leaves it behind. A reinstall into the same namespace inherits a token
minted against the previous PMM's Grafana database, every pmm-client sidecar in the bundled
PostgreSQL cluster fails to register, and the cluster is silently unmonitored:

```
Failed to register pmm-agent on PMM Server: Auth method is not service account token
```

Measured on tibi-test: **0 successful registrations against 159 auth failures**, and no
PostgreSQL service, exporter or QAN agent for the bundled cluster in PMM's inventory. Nothing
surfaces it — `PMM_AGENT_SIDECAR` retries inside the entrypoint, so the container never
restarts and the pod stays Ready.

Introduced by PMM-14408 (d161f5f).

## Fix

- The Job probes the stored token against `/graph/api/user` and re-mints on any non-200. The
  readiness loop moves above that check, because probing an unready PMM would re-mint a token
  that was fine. The write path already upserts.
- The secret name is release-scoped. A new `pmm.pgPmmSecretName` helper mirrors pg-db's own
  default and resolves to `<release>-pg-db-pmm-secret`; the Job and the pg-db subchart both
  derive from it so they cannot drift. `pg-db.pmm.secret` is empty by default and still
  honoured when set, for adopting an existing secret.

## Verified

Live on tibi-test, all three paths:

| Path | Result |
|---|---|
| secret absent | created `pmm2-pg-db-pmm-secret` |
| valid token | `authenticates (HTTP 200). Skipping token creation.` — token byte-identical, no churn on upgrade |
| stale token | `does NOT authenticate (HTTP 401) - re-minting.` — on current `PMM-HA-GA` the same input logs "Skipping" and leaves PostgreSQL unmonitored |

Naming is consistent end to end — Job, `PostgresCluster` and all 3 instance StatefulSets
resolve to the same secret, and an explicit override still wins in both. `postgres_exporter`
agents went 1 → 5 and auth failures 159 → 0.

## Notes for review

- **Upgrade timing.** The operator takes several minutes to reconcile the rename into the
  StatefulSets; until then pods reference the old name. Don't delete `pg-pmm-secret` during the
  upgrade — let it become orphaned and remove it after.
- **No `Chart.yaml` bump** here, to avoid adding another version conflict on top of #936/#937.
  Happy to add one if preferred.
- `docs/pmm-backup.md` tells you to fix this by deleting the Job and letting Helm recreate it,
  which hits the same skip — the secret must be deleted. That file only exists on #919, so the
  correction belongs there.


charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -405,3 +405,22 @@ dict with "name" and "value".
...
+by the token-init Job rather than owned by Helm, so `helm uninstall` cannot remove it. A
+fixed name therefore lets a NEW release inherit the previous install's dead token. The Job
+and the pg-db subchart must agree on this name, so both derive it from here.
+*/}}
+{{- define "pmm.pgPmmSecretName" -}}
+{{- $explicit := index .Values "pg-db" "pmm" "secret" -}}
+{{- if $explicit -}}
+{{- $explicit -}}
+{{- else -}}
+{{- printf "%s-pg-db-pmm-secret" .Release.Name -}}
```
ademidoff: This rebuilds pg-db's default name by hand rather than sharing a source of truth with it. pg-db renders `{{ .Values.pmm.secret | default (printf "%s-pmm-secret" (include "pg-database.fullname" .)) }}` (`charts/pg-db/templates/cluster.yaml:467`), so the `-pg-db-` segment here is a *copy* of the `pg-database.fullname` override at line 91, not a reference to it.

That makes line 417 -- "The Job and the pg-db subchart must agree on this name, so both derive it from here" -- untrue as written: the subchart derives it from `pg-database.fullname`, only the Job derives it from here. And it means the override at line 91 and this literal have to be edited together, which is the drift the helper exists to prevent.

```suggestion
{{- printf "%s-pmm-secret" (include "pg-database.fullname" .) -}}
```

Verified: with that substitution `helm template` output is identical (apart from the chart's randomly generated passwords/certs) for the default, for `--set pg-db.pmm.secret=adopted`, and for a 53-character release name -- the parent's `pg-database.fullname` override does win over the subchart's definition, so it still resolves to `<release>-pg-db-pmm-secret`.

ademidoff (commented):
Rendered `charts/pmm-ha` at 8105343 (default, `--set pg-db.pmm.secret=...`, and a 53-character release name) and extracted the rendered Job script to run every branch of the new check against stubbed `curl`/`kubectl`: the name agrees in both directions and the 200/401 paths behave as you describe. Look at the re-mint trigger first -- a curl failure (`000`) or a transient 5xx takes the same branch as a 401. On the `Chart.yaml` question in your notes: leaving it out is right for the PMM charts, they accumulate and get bumped once in the release PR.

charts/pmm-ha/values.yaml
```diff
@@ -836,10 +836,15 @@ pg-db:
     image:
       repository: docker.io/percona/pmm-client
       tag: "3"
-    ## @param pg-db.pmm.secret Name of the secret containing PMM_SERVER_TOKEN
-    ## This secret is automatically created by the pmm-token-init Job
-    ## The secret must contain a key named PMM_SERVER_TOKEN with the service account token
-    secret: pg-pmm-secret
+    ## @param pg-db.pmm.secret Name of the secret containing PMM_SERVER_TOKEN. Leave empty to derive "<release>-pg-db-pmm-secret"
+    ## Created automatically by the pmm-token-init Job, which must agree with pg-db on the
+    ## name - both derive it from the pmm.pgPmmSecretName helper when this is unset.
```
ademidoff: Same inaccuracy as the helper comment: pg-db does not derive the name from `pmm.pgPmmSecretName`. Its own template falls back to `printf "%s-pmm-secret" (include "pg-database.fullname" .)`, and the helper mirrors that string.

```suggestion
    ## name - the Job takes it from the pmm.pgPmmSecretName helper, which resolves to the same
    ## "<pg-database.fullname>-pmm-secret" that pg-db falls back to when this is unset.
```

charts/pmm-ha/templates/pg-pmm-token-job.yaml
```diff
@@ -115,6 +106,32 @@ spec:
...
+            EXISTING_KEY=$(kubectl get secret "$SECRET_NAME" -n "$NAMESPACE" -o jsonpath='{.data.PMM_SERVER_TOKEN}' 2>/dev/null | base64 -d)
+            if [ -n "$EXISTING_KEY" ]; then
+              TOKEN_CODE=$(curl -sk -o /dev/null -w "%{http_code}" \
+                -H "Authorization: Bearer ${EXISTING_KEY}" \
+                "${PMM_URL}/graph/api/user" 2>/dev/null) || true
+              if [ "$TOKEN_CODE" = "200" ]; then
+                echo "Secret '$SECRET_NAME' holds a token that authenticates (HTTP $TOKEN_CODE). Skipping token creation."
+                exit 0
+              fi
+              echo "Secret '$SECRET_NAME' holds a token that does NOT authenticate (HTTP $TOKEN_CODE) - re-minting."
```
ademidoff, line 160: `!= 200` is being read as "the token is dead", but only 401/403 say anything about the token. I pulled the rendered script out of the Job and ran it with stubbed `kubectl`/`curl`:

```
HTTP 401                        -> "does NOT authenticate (HTTP 401) - re-minting."   correct
HTTP 503                        -> "does NOT authenticate (HTTP 503) - re-minting."   wrong
curl exit 7, %{http_code}=000   -> "does NOT authenticate (HTTP 000) - re-minting."   wrong
```

For `000` the request never completed, so the message on line 131 is itself false in that case.

The comment on 119-120 says running below the readiness loop covers this, and it does cover a cold PMM -- but the loop only proves `/v1/readyz` answered 200 *once*, through HAProxy, whose backend check is `/v1/server/leaderHealthCheck` (`charts/pmm-ha/templates/haproxy-configmap.yaml`). A leader change, or a backend going through re-check, between those two requests lands here.

A false positive is not free: the secret gets a fresh token while running PG pods keep the old one from `secretKeyRef` until they restart, and the previous service account token is never revoked in Grafana, so each occurrence leaves one behind. `restartPolicy: OnFailure` + `backoffLimit: 10` is already the right mechanism for an inconclusive probe:

```suggestion
              if [ "$TOKEN_CODE" = "200" ]; then
                echo "Secret '$SECRET_NAME' holds a token that authenticates (HTTP $TOKEN_CODE). Skipping token creation."
                exit 0
              fi
              if [ "$TOKEN_CODE" != "401" ] && [ "$TOKEN_CODE" != "403" ]; then
                echo "ERROR: could not validate the token in '$SECRET_NAME' (HTTP $TOKEN_CODE). Not re-minting on an inconclusive probe; failing so the Job retries."
                exit 1
              fi
              echo "Secret '$SECRET_NAME' holds a token that does NOT authenticate (HTTP $TOKEN_CODE) - re-minting."
```

charts/pmm-ha/templates/pg-pmm-token-job.yaml
```diff
@@ -63,7 +63,7 @@ spec:
         - name: PMM_PORT
           value: "443"
         - name: SECRET_NAME
-          value: "{{ index .Values "pg-db" "pmm" "secret" | default "pg-pmm-secret" }}"
+          value: "{{ include "pmm.pgPmmSecretName" . }}"
```
ademidoff, line 83: Worth settling before this lands on the GA branch: `spec.template` on a Job is immutable, and this Job carries no `helm.sh/hook` annotations, so Helm treats it as an ordinary release resource and patches it on upgrade. This PR changes both this env value and the `command` below, so a `helm upgrade` from a release whose `<release>-pmm-token-init` Job still exists is rejected with `cannot patch "<release>-pmm-token-init" with kind Job: Job.batch "..." is invalid: spec.template: field is immutable`. `ttlSecondsAfterFinished: 86400` (line 46) bounds it to the 24h after the Job completed, and CI only does a fresh `ct install`, so nothing here will surface it. Reasoned, not reproduced -- no cluster on hand.

The sibling `pmm-secret-init-job.yaml` sidesteps this with `helm.sh/hook: pre-install` plus `helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded`. Copying that verbatim is not enough here: hooks block the release, and the readiness loop on 87-107 is `while true` with no deadline, so a `post-install,post-upgrade` hook would hang `helm upgrade` indefinitely whenever PMM never becomes ready. Either bound that loop and make it a hook, or fold a hash of the pod template into the Job name. Fine as a follow-up if you would rather not grow this PR -- flagging it because the rename is what makes the patch non-empty, so it is this PR's own upgrade path that trips.

charts/pmm-ha/README.md
```diff
@@ -306,11 +306,18 @@ pmm-admin config --server-url=https://service_token:<token>@pmm-ha-haproxy:443 -
...
 3. PostgreSQL pods use this token to authenticate and push metrics to PMM via the `pmm-ha-haproxy` endpoint
 
 No manual configuration is required - the Percona PostgreSQL Operator handles the integration automatically.
 
+The Job creates that secret with `kubectl`, so Helm does not own it and `helm uninstall`
+leaves it behind. On every run the Job checks whether the stored token still authenticates
+and re-mints it if not, so a reinstall into a namespace that still holds the old secret
+recovers on its own. If PostgreSQL monitoring is missing, check the `pmm-client` container
+in a PostgreSQL pod: `Auth method is not service account token` means the token was rejected,
+and the Job's log will say whether it re-minted.
```
ademidoff: The rename is a behaviour change for existing installs, and this section is where a user would look for it. Your PR description already has the operational detail ("the operator takes several minutes to reconcile the rename into the StatefulSets ... don't delete `pg-pmm-secret` during the upgrade -- let it become orphaned and remove it after") -- that belongs here rather than only in the PR:

```suggestion
and the Job's log will say whether it re-minted.

Upgrading from a chart version that used the old fixed `pg-pmm-secret` name: after the upgrade the
Job creates `<release>-pg-db-pmm-secret`, and the PostgreSQL operator needs a few minutes to
reconcile the new name into the instance StatefulSets. Leave `pg-pmm-secret` in place until the
PostgreSQL pods have restarted against the new secret, then delete it.
```

charts/pmm-ha/templates/_helpers.tpl
theTibi: Taken as suggested in b28c338. You are right that the comment was untrue as written — only the Job read the helper. Both sides now go through `pg-database.fullname`, so the override at line 91 is the single place to edit.

charts/pmm-ha/values.yaml
theTibi: Fixed in b28c338, same correction.

charts/pmm-ha/templates/pg-pmm-token-job.yaml
theTibi: Applied in b28c338.

Reproduced live rather than with a stub: an nginx proxy in front of the real PMM that passes everything through except `/graph/api/user`, which returns 503. Readiness passes, then the token probe fails.

- Previous code: `does NOT authenticate (HTTP 503) - re-minting.` — a healthy token was replaced.
- This commit: `ERROR: could not validate the token ... Not re-minting on an inconclusive probe; failing so the Job retries.` Secret untouched.

Removing the 503 override while the Job was still retrying, it recovered on its own and logged `authenticates (HTTP 200). Skipping token creation.`, so failing closed does not strand it. 401 and 200 paths re-checked against the real PMM and unchanged.

charts/pmm-ha/templates/pg-pmm-token-job.yaml
theTibi: Confirmed on a cluster, and fixed here rather than deferred.

Installing the pre-change chart and upgrading to this PR gives exactly what you predicted, and the release is left `failed`:

```
Error: UPGRADE FAILED: cannot patch "tk-pmm-token-init" with kind Job:
Job.batch "tk-pmm-token-init" is invalid: spec.template: ... field is immutable
```

Took the second option — the Job name now carries a hash of its own pod template, so a template change renames the resource and Helm creates the new Job and prunes the old one. The same upgrade on b28c338 succeeds, and a repeat upgrade with no changes is a no-op. Not the hook route, for the reason you gave: the readiness loop is `while true` with no deadline.

One thing that came out of testing it: the old name was already broken at long release names. `spec.template.labels` carries the Job name and label values cap at 63 bytes, so a 53-character release fails at install time.

```
Job.batch "zzz...-pmm-token-init" is invalid: spec.template.labels:
Invalid value: "zzz...-pmm-token-init": must be no more than 63 bytes
```

The helper truncates the release name rather than the finished string, so that case now installs.

charts/pmm-ha/README.md
theTibi: Added in b28c338, along with a note on the inconclusive-probe behaviour and how to find the Job now that its name ends in a hash.

JiriCtvrtka (approved):

PR #944, opened 2026-09-02
PMM-15400 Set the Grafana admin password from a pre-created pmm-secret

theTibi:
## Problem

`charts/pmm-ha` maps the secret's `PMM_ADMIN_PASSWORD` to Grafana's
`GF_SECURITY_ADMIN_PASSWORD` only inside `{{ if .Values.secret.create }}`.
`secret.create` defaults to **false** and the documented install pre-creates
`pmm-secret`, so on the documented path Grafana never receives the password — it
stays at the built-in `admin/admin`. Nothing in the PMM server image consumes
`PMM_ADMIN_PASSWORD` (`managed/utils/envvars/parser.go` skips it), so the
documented key has no effect at all there.

Knock-on effect: the chart's own `pmm-token-init` job authenticates as
`admin:${PMM_ADMIN_PASSWORD}`, gets 401 and ends in CrashLoopBackOff, so the
PostgreSQL monitoring service account is never created.

`charts/pmm` guards the same env on `secret.name` and has since PMM-13679
(fd11e06); `pmm-ha` was branched from the PoC and kept the pre-fix form.

## Fix

- Emit `GF_SECURITY_ADMIN_PASSWORD` whenever `secret.name` is set. `envFrom`
  injects the secret's keys verbatim and cannot rename one, so the mapping is
  needed on both paths.
- Mark the ref required, so a missing key fails visibly instead of silently
  falling back to the built-in password.
- Add `pmm.validateSecret`: the PMM container mounts seven keys from `pmm-secret`
  with no `optional`, so one missing key left every pod in
  `CreateContainerConfigError` naming nothing. All missing keys are now reported
  in one message. Skipped under `secret.create` — that list is exactly what
  `secret.yaml` generates.
- Docs: `PMM_ADMIN_PASSWORD` is the web UI / API admin password and is required
  with a pre-created secret; `GF_PASSWORD` is the Grafana PostgreSQL user, not the
  UI login.

## Verified

Fresh install on EKS, pre-created secret, `secret.create: false`:

| Check | Result |
|---|---|
| `/v1/ha/status` as `admin:<secret password>` | 200 |
| `/v1/ha/status` as `admin:admin` | 401 |
| Same through HAProxy | 200 |
| `pmm-token-init` job | Completed, `pg-pmm-secret` created |
| secret missing the key | install fails naming `PMM_ADMIN_PASSWORD` |
| no secret at all | unchanged "Secret not found" message |

No `Chart.yaml` bump, matching #941 and this branch's release-time bump convention.


charts/pmm-ha/templates/statefulset.yaml
```diff
@@ -172,6 +173,31 @@ spec:
...
+
+            Where envFrom does apply, env wins over it, so a GF_SECURITY_ADMIN_PASSWORD key also
+            placed in the secret is overridden - PMM_ADMIN_PASSWORD is the single source of truth.
+
+            Required, not optional: pmm.validateSecret refuses to render without the key, so an
+            optional ref would only serve to let a key deleted after install silently drop the
+            env var again. Note charts/pmm still uses optional: true for its copy of this ref.
+            */}}
+            {{ if .Values.secret.name }}
+            - name: GF_SECURITY_ADMIN_PASSWORD
```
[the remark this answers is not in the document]
theTibi: CI passed without a bump. `ct` prints `Version increment checking disabled.` — the check is inert when charts are named explicitly via `--charts=…`, which this workflow does, and `.github/ct.yaml` sets `target-branch: main`, so the comparison base would be wrong on a release branch anyway. On `PMM-HA-GA` the version is bumped by the release commits; chart-releaser only runs on push to `main`. #941 likewise touches no `Chart.yaml`.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -84,6 +84,51 @@ checksum/clickhouse-datasource: {{ include (print $.Template.BasePath "/clickhou
...
+demanding a key the chart is about to write would abort the install on its own output.
+
+PMM_ADMIN_PASSWORD is the key PMM-15400 is about - statefulset.yaml maps it to Grafana's
+GF_SECURITY_ADMIN_PASSWORD, and leaving that ref optional is what let it vanish silently.
+
+Included from statefulset.yaml, the template that depends on these keys.
+*/}}
+{{- define "pmm.validateSecret" -}}
+{{- if not .Values.secret.create -}}
+{{- $found := lookup "v1" "Secret" .Release.Namespace .Values.secret.name -}}
```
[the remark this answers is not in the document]
theTibi: Right principle, wrong templates. Helm sorts same-depth templates in reverse alphabetical order, so `statefulset.yaml` renders *before* both `pg-user-credentials-secrets.yaml` and `clickhouse-cluster.yaml`. A secret missing `PMM_ADMIN_PASSWORD`, `GF_PASSWORD` and `PMM_CLICKHOUSE_USER` fails once, naming all three.

It does hold for `vmauth.yaml`, which renders before `statefulset.yaml` and does an unguarded `b64dec` — a missing `VMAGENT_remoteWrite_basicAuth_*` died with `invalid value; expected string`, naming neither key nor secret. Added the include there in 8080d38.

Still not a complete guard: it is a no-op under `secret.create: true`, where the `b64dec` fails the same way, and `clickhouse-cluster.yaml` reads a key the same way. Fixing that at the read site rather than per consumer is PMM-15436.

charts/pmm-ha/README.md
```diff
@@ -471,7 +471,10 @@ Percona will release a new chart updating its containers if a new version of the
 
 PMM admin password would be set only on the first deployment. That setting is ignored if PMM was already provisioned and just restarted and/or updated. In real-life situations it is recommended to create the `pmm-secret` secret manually before the release and set `secret.create` to false. The chart then won't overwrite secret during install or upgrade and values.yaml won't contain any secret.
 
-If PMM admin password is not set explicitly (default), it will be generated.
+If PMM admin password is not set explicitly, it will be generated - but only when
+`secret.create: true`. With a manually created `pmm-secret` (the default), `secret.pmm_password`
+is ignored and the secret's `PMM_ADMIN_PASSWORD` key is required: the chart fails the install if
+it is missing, because without it PMM would silently keep its built-in admin password.
```
ademidoff, line 477: The install-time story is right, but the upgrade path is missing, and that is the path an operator already hit by PMM-15400 is on. Two separate things happen there.

**1. The gate applies to `helm upgrade` too, not just install.** `pmm.validateSecret` is not gated on `.Release.IsInstall`, so a release whose pre-created secret has no `PMM_ADMIN_PASSWORD` now fails to *upgrade*, at render. Reproduced against a live cluster with a real `lookup`:

```
$ helm template pmm-ha charts/pmm-ha -n <ns> --dry-run=server --set secret.name=<secret-without-the-keys>
Error: execution error at (pmm-ha/templates/vmauth.yaml:10:4): Secret '<...>' in namespace '<ns>' is
missing, or has an empty value for, required key(s): PMM_ADMIN_PASSWORD, GF_PASSWORD, PG_PASSWORD, ...
```

That is the right outcome, but "fails the install" here and "refuses to install" at line 525 both read, to someone with a running release, as "does not apply to me".

**2. Adding the key and upgrading still does not set the password.** Grafana applies `GF_SECURITY_ADMIN_PASSWORD` only when it creates the `admin` user. PMM's own docs say so — `reference/faq.md`: "set the password for the default `admin` user *on first startup*", and the 3.8.0 note for PMM-14934 repeats "at first start" — and nothing in this chart invokes PMM's `change-admin-password`. So on an already-provisioned deployment the UI stays on `admin/admin` and `<release>-pmm-token-init` keeps getting 401: the knock-on effect in the PR description is fixed for fresh installs only.

The unchanged sentence above ("PMM admin password would be set only on the first deployment...") is technically the caveat, but next to the new paragraph it now reads as being about `secret.pmm_password`. One sentence closes it:

```suggestion
it is missing, because without it PMM would silently keep its built-in admin password.

For a release that was already installed without the key, adding it and upgrading is not enough:
Grafana applies `GF_SECURITY_ADMIN_PASSWORD` only when it first creates the `admin` user, so the
password has to be reset out of band (`change-admin-password` in the PMM container, or the Grafana
API) before `<release>-pmm-token-init` stops getting 401.
```

Confidence: high on the mechanism (PMM's own documentation, plus no reset step anywhere in the chart); I did not run an actual upgrade of a pre-fix release.

ademidoff (commented):
Verified by rendering rather than reading: with a real `lookup` (`--dry-run=server` against a live cluster) `pmm.validateSecret` fires from `vmauth.yaml` first and names all seven keys in one message; `GF_SECURITY_ADMIN_PASSWORD` is emitted exactly once under both `secret.create` values and there is no `envFrom` secretRef when it is true, as the comment claims; and the generated password appears nowhere in the render outside the Secret itself. The upgrade path is the one thing worth a second look — see the comment on the README hunk.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -84,6 +84,53 @@ checksum/clickhouse-datasource: {{ include (print $.Template.BasePath "/clickhou
...
+
+PMM_ADMIN_PASSWORD is the key PMM-15400 is about - statefulset.yaml maps it to Grafana's
+GF_SECURITY_ADMIN_PASSWORD, and leaving that ref optional is what let it vanish silently.
+
+Included from statefulset.yaml and vmauth.yaml - both read these keys, and vmauth.yaml
+renders first, so it needs its own call to report the missing key rather than dying on a
+b64dec. Keep every consumer that decodes a key from this secret calling it.
+*/}}
+{{- define "pmm.validateSecret" -}}
+{{- if not .Values.secret.create -}}
```
ademidoff, line 105: Minor, but verified: this guard lets an empty `secret.name` through, and `lookup` with an empty name does not come back empty — it returns a `SecretList`, which is truthy and has no `.data`, so every key is reported missing against a secret that was never named.

```
$ helm template pmm-ha charts/pmm-ha -n <ns> --dry-run=server --set secret.name=""
Error: execution error at (pmm-ha/templates/vmauth.yaml:10:4): Secret '' in namespace '<ns>' is missing,
or has an empty value for, required key(s): PMM_ADMIN_PASSWORD, GF_PASSWORD, PG_PASSWORD, ...
```

`statefulset.yaml` already treats an empty `secret.name` as "no secret" — both for `envFrom` (line 155) and for the new `GF_SECURITY_ADMIN_PASSWORD` ref (line 193) — so reading it here as "a secret with seven missing keys" is inconsistent with the file that calls this.

```suggestion
{{- if and (not .Values.secret.create) .Values.secret.name -}}
```

The install fails either way (plenty of other refs in this chart use `secret.name` unguarded), so this is purely about which message the operator gets.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -84,6 +84,53 @@ checksum/clickhouse-datasource: {{ include (print $.Template.BasePath "/clickhou
...
+{{/*
+Only inspect keys once the Secret is actually in hand. `lookup` also comes back empty on every
+client-side render - helm template, --dry-run=client, a GitOps preview - where it says nothing
+about the secret's contents, and reporting all seven keys as missing there would be simply
+wrong. When the secret really is absent, pg-user-credentials-secrets.yaml fails with the
+accurate "Secret not found" message instead.
+*/}}
+{{- if $found -}}
+{{- $data := $found.data | default dict -}}
+{{- $required := list "PMM_ADMIN_PASSWORD" "GF_PASSWORD" "PG_PASSWORD" "PMM_CLICKHOUSE_USER" "PMM_CLICKHOUSE_PASSWORD" "VMAGENT_remoteWrite_basicAuth_username" "VMAGENT_remoteWrite_basicAuth_password" -}}
```
ademidoff, line 116: `GF_PASSWORD` and `PG_PASSWORD` are now validated twice: `pg-user-credentials-secrets.yaml:8-13` still carries its own per-key `fail` for exactly these two. Because same-depth templates render in reverse alphabetical order, `statefulset.yaml` reaches this helper first, so on the documented `secret.create: false` path those two checks can no longer fire — the operator will only ever see the combined message from here.

They stay reachable for `secret.create: true`, but that path has its own problem: `pg-user-credentials-secrets.yaml:3` does an unconditional render-time `lookup`, which on a fresh install runs before the `pre-install` hook in `secret.yaml` has created anything, so it fails with "Secret not found" regardless (pre-existing, not from this PR).

Since this change makes this helper the single place that reports a missing key, dropping lines 8-13 of `pg-user-credentials-secrets.yaml` in the same commit would leave one message per missing key instead of two that cannot both be reached.

charts/pmm-ha/README.md
```diff
@@ -356,7 +356,7 @@ To create additional service tokens manually, see the [PMM documentation on serv
...
 | `secret.annotations`  | Defines the annotations of the k8s secret that holds passwords and other secrets                                                                                                   | `{}`         |
 | `secret.create`       | If true then secret will be generated by Helm chart. Otherwise it is expected to be created by user and all of its keys will be mounted as env vars.                                                                               | `false`       |
-| `secret.pmm_password` | Initial PMM password - it changes only on the first deployment, ignored if PMM was already provisioned and just restarted. If PMM admin password is not set, it will be generated. | `""`         |
+| `secret.pmm_password` | Initial PMM password - it changes only on the first deployment, ignored if PMM was already provisioned and just restarted. If PMM admin password is not set, it will be generated. Only used when secret.create is true; with a manually created secret the PMM_ADMIN_PASSWORD key is required instead. | `""`         |
```
ademidoff, line 359: `values.yaml` qualified this sentence and the table row did not, so the two now disagree on the exact point the PR is making. The `@param` line at `values.yaml:202` drops "If PMM admin password is not set, it will be generated" and re-states it below as "- only when `secret.create` is true", while this row keeps it unconditional, one clause before "Only used when secret.create is true".

`CLAUDE.md`: "`values.yaml` uses `## @section` / `## @param` annotations, and each chart's `README.md` carries a hand-maintained parameter table generated from them."

```suggestion
| `secret.pmm_password` | Initial PMM password - it changes only on the first deployment, ignored if PMM was already provisioned. Only used when secret.create is true; with a manually created secret the PMM_ADMIN_PASSWORD key is required instead. | `""`         |
```

charts/pmm-ha/README.md
theTibi: Taken as written in b908542, plus the same widening in the Note below and in `values.yaml`.

Confirmed both halves: no `.Release.IsInstall` guard, and nothing in the image resets the password — `entrypoint.sh` never calls `change-admin-password` (only the AMI role does), and `managed/utils/envvars/parser.go:227` skips `PMM_ADMIN_PASSWORD`.

charts/pmm-ha/templates/_helpers.tpl
theTibi: Right about the `SecretList`, but that one-liner makes the message worse: skipping the loop just moves the failure one template on, and `--set secret.name=""` then dies at `vmauth.yaml:17` with `index of untyped nil`, naming neither the secret nor the setting.

b908542 fails on it explicitly instead: `secret.name is empty. Set it to the name of the Kubernetes Secret that holds the PMM credentials (the chart default is 'pmm-secret')`. The other messages are unchanged.

charts/pmm-ha/templates/_helpers.tpl
theTibi: Agreed — dropped in #964 (2642a5b). With #944 in, a secret missing `GF_PASSWORD` fails once, at `vmauth.yaml:10`, with the combined message. #964 notes it has to merge second.

Your `pg-user-credentials-secrets.yaml:3` note was the bigger find: that unconditional `lookup` means `secret.create: true` has never been able to do a fresh install — it aborts before `secret.yaml`'s own pre-install hook runs. Filed as PMM-15501, fixed in #964, verified with a real install and upgrade on EKS.

charts/pmm-ha/README.md
theTibi: Taken as written in b908542 — the row and the `@param` at `values.yaml:202` now match exactly.

JiriCtvrtka (approved):

PR #948, opened 2026-09-04
[psmdb-db] render metadata.finalizers only when the list is non-empty

KaivalyaMDabhadkar:
When `finalizers` is set to an empty list, the chart renders `metadata.finalizers: []` on the PerconaServerMongoDB resource. The API server drops the empty list, so the live object never matches the rendered manifest and GitOps tools such as ArgoCD report the resource as permanently OutOfSync.

This change wraps the block in `{{- if .Values.finalizers }}`, the same way `pg-db` already does, so the key is only rendered when there is something in it. The default value is unchanged and still renders `percona.com/delete-psmdb-pods-in-order`.

Testing:
- `helm template` with default values produces byte-identical output to `main`.
- `helm template --set-json 'finalizers=[]'` (and `null`) no longer emits `metadata.finalizers`. Custom lists still render.
- `helm lint charts/psmdb-db` passes.
- Verified with Helm 3.16.3, 3.17.1, 3.19.2 and 4.2.4.

### DIFF
--- a/charts/psmdb-db/Chart.yaml
+++ b/charts/psmdb-db/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "1.23.0"
 description: A Helm chart for installing Percona Server MongoDB Cluster Databases using the PSMDB Operator.
 name: psmdb-db
 home: https://www.percona.com/doc/kubernetes-operator-for-psmongodb/index.html
-version: 1.23.0
+version: 1.23.1
 maintainers:
   - name: hors
     email: slava.sarzhan@percona.com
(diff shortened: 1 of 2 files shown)


lalitadithya (approved):
thanks for this change, we have been using this patch it is working as intended for us! :)

jvpasinatto:
Hi @KaivalyaMDabhadkar could you also bump the `psmdb-db` chart version to `1.23.1`? Then we can merge. Thanks!

KaivalyaMDabhadkar:
> Hi @KaivalyaMDabhadkar could you also bump the `psmdb-db` chart version to `1.23.1`? Then we can merge. Thanks!

Done, bumped the chart version to 1.23.1. Thanks!

jvpasinatto (approved):

valmiranogueira (approved):

jvpasinatto:
Merged. @KaivalyaMDabhadkar Thanks for the contribution!

PR #950, opened 2026-09-04
PMM-15426 Make the pmm-ha chart installable on OpenShift

theTibi:
## Problem

On OpenShift with `restricted-v2`, `helm install charts/pmm-ha` reports `STATUS: deployed` while four workloads are rejected at admission and produce **zero pods**, and HAProxy crash-loops for two further reasons.

Verified live on ROSA HCP, OpenShift 4.21.30, from `PMM-HA-GA` @ 924116e.

| # | Workload | Cause |
|---|---|---|
| 1 | PMM Server StatefulSet | `podSecurityContext` `runAsUser: 1000` / `fsGroup: 1000` — outside the namespace's assigned ranges |
| 2 | PMM Client StatefulSet | hardcoded `fsGroup: 1002`, no values knob, so no user-side workaround existed |
| 3 | kube-state-metrics | subchart pins uid/gid/fsGroup `65534` |
| 4 | prometheus-node-exporter | `hostNetwork`, `hostPID`, hostPath volumes, host port 9100 |
| 5 | HAProxy | resolver named `kube-dns.kube-system`, which does not exist on OpenShift (`dns-default.openshift-dns`); HAProxy rejects an unresolvable nameserver at **config-parse time** |
| 6 | HAProxy | bound `*:443` while `restricted-v2` drops `NET_BIND_SERVICE`: `cannot bind socket (Permission denied) for [0.0.0.0:443]` |

HAProxy is the only ingress to PMM, so 5 and 6 keep PMM unreachable even once the SCC issues are fixed.

ClickHouse, VictoriaMetrics, PostgreSQL and the three operators are **not** affected — they admit cleanly under `restricted-v2`.

## Changes

- New top-level `openshift` value gating both pod securityContexts, plus a `pmmClient.fsGroup` knob. `runAsUser` was never load-bearing — the PMM Server image already declares `USER 1000` and its entrypoint handles an arbitrary assigned uid via the NSS wrapper. `fsGroup` is the part that matters, and OpenShift assigns one when the chart asks for none.
- `haproxy.containerPorts.https` now drives the HAProxy bind and every internal reference to it (`pmm.client.serverAddress`, `pg-pmm-token-job`'s `PMM_PORT`), since the subchart derives the Service port from the same value.
- HAProxy resolver uses `parse-resolv-conf`, so one line works on both platforms.
- New `examples/values-openshift.yaml` covering all six.
- README: an **Installing on OpenShift** section, the two new parameters, and a correction — `podSecurityContext`'s default was documented as `{}` while the chart ships `runAsUser: 1000, fsGroup: 1000`.

## Verification

**No regression on plain Kubernetes** — with default values every workload's pod `securityContext` renders byte-identical to before, and HAProxy still binds and publishes 443. (This held only after review: the first version of the helper emitted a whitespace-only line ahead of `securityContext:`. Fixed, and a CI render check now pins it.)

**The bundled PostgreSQL reaches PMM** — verified on plain Kubernetes with `haproxy.containerPorts.https: 8443`. With the bare `pg-db.pmm.serverHost` the sidecar looped on `dial tcp …:443: i/o timeout` while the pod stayed 5/5 Running and the inventory stayed empty; with the port spelled out it registers and `postgres_exporter` runs.

**On OpenShift**, with `-f examples/values-openshift.yaml`, the full stack reaches green: PMM Server 3/3, PMM Client 3/3, HAProxy 3/3, ClickHouse 3+3, VictoriaMetrics all up, PostgreSQL 3 + pgbouncer, kube-state-metrics 1/1, `pg-pmm-secret` created. PMM Server runs under the assigned identity with a writable volume:

```
uid=1000860000 gid=0(root) groups=0(root),1000860000
drwxrwsr-x. 14 root 1000860000 /srv     → writable
```

Full reproduce/verify commands are in the ticket.

## Not in this PR

Three bugs surfaced during verification that are **not** OpenShift-specific and need their own tickets:

- `pg-pmm-token-job.yaml` is a plain Job with a fixed name and no Helm hook, so any values change touching its pod spec fails `helm upgrade` with `spec.template: field is immutable`.
- A `%` in any password breaks supervisord — `!` URL-encodes to `%21`, ConfigParser reads `%` as a format char and refuses `vmalert.ini`, so vmalert/grafana/qan-api2/vmproxy never start and readyz stays 504.
- `PMM_ADMIN_PASSWORD` is never applied; pmm-init's only password task is skipped for non-AMI, so the admin password stays `admin` and both `pg-pmm-token-job` and `pmm-client-register` get 401.

https://perconadev.atlassian.net/browse/PMM-15426

### DIFF
--- a/charts/pmm-ha/Chart.yaml
+++ b/charts/pmm-ha/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: pmm-ha
 description: A Helm chart for Percona Monitoring and Management (PMM)
 type: application
-version: 1.6.2
+version: 1.6.3
 appVersion: "3.9.1"
 home: https://github.com/percona/pmm
 maintainers:
--- a/charts/pmm-ha/README.md
+++ b/charts/pmm-ha/README.md
@@ -387,6 +387,7 @@ To create additional service tokens manually, see the [PMM documentation on serv
 | ----------------------------- | --------------------------------------------------------------------------------------------------------------- | ----------- |
 | `haproxy.service.type`        | Service type for HAProxy: ClusterIP (internal), LoadBalancer (external via LB), or NodePort (external via node) | `ClusterIP` |
 | `haproxy.service.annotations` | Service annotations (add cloud-specific annotations as needed)                                                   | `{}`        |
+| `haproxy.containerPorts.https` | Port HAProxy binds for PMM traffic; the Service publishes the same port. Must be above 1024 on OpenShift        | `443`       |
 
 
 ### PMM storage configuration
@@ -410,7 +411,8 @@ To create additional service tokens manually, see the [PMM documentation on serv
 | `serviceAccount.annotations` | Annotations for service account. Evaluated as a template. Only used if `create` is `true`.                          | `{}`                  |
 | `serviceAccount.name`        | Name of the service account to use. If not set and create is true, a name is generated using the fullname template. | `pmm-service-account` |
 | `podAnnotations`             | Pod annotations                                                                                                     | `{}`                  |
-| `podSecurityContext`         | Configure Pods Security Context                                                                                     | `{}`                  |
+| `podSecurityContext`         | Configure Pods Security Context. Ignored when `openshift` is `true`                                                 | `{runAsUser: 1000, fsGroup: 1000}` |
+| `openshift`                  | Set to `true` on OpenShift so the chart emits no pod securityContext and the cluster assigns the uid/fsGroup        | `false`               |
 | `securityContext`            | Configure Container Security Context                                                                                | `{}`                  |
 | `nodeSelector`               | Node labels for pod assignment                                                                                      | `{}`                  |
 | `tolerations`                | Tolerations for pod assignment                                                                                      | `[]`                  |
@@ -655,6 +657,21 @@ victoriaMetrics:
     extraArgs:
       maxLabelsPerTimeseries: "60"
 ```
+
+### Installing on OpenShift
+
+OpenShift's `restricted-v2` SCC gives every namespace its own uid and supplemental-group ranges
+and rejects any pod asking for values outside them, and it runs containers without
+`NET_BIND_SERVICE`. Install with the bundled overlay, which covers all of it:
+
+```bash
+helm install pmm-ha percona/pmm-ha -n pmm -f examples/values-openshift.yaml
+```
+
+It sets `openshift: true` (PMM Server and PMM Client let the cluster assign uid and fsGroup),
+disables the bundled node-exporter in favour of OpenShift's, turns off the kube-state-metrics
+securityContext, and moves the HAProxy port to 8443. Publish 443 externally with a Route or
+Ingress pointing at the HAProxy Service.
 
 ### Using OpenShift's node exporter
 
--- a/charts/pmm-ha/examples/values-openshift.yaml
+++ b/charts/pmm-ha/examples/values-openshift.yaml
@@ -1,0 +1,33 @@
+## OpenShift overlay for the pmm-ha chart.
+##
+##   helm install pmm-ha percona/pmm-ha -n <namespace> -f examples/values-openshift.yaml
+##
+## OpenShift's `restricted-v2` SCC gives every namespace its own uid range and supplemental-group
+## range and rejects any pod requesting values outside them. Each setting below clears one such
+## rejection; without them the StatefulSets and Deployments are admitted-then-refused and Helm
+## still reports STATUS: deployed while no pods exist.
+
+## PMM Server and PMM Client: emit no pod securityContext, so OpenShift assigns uid and fsGroup.
+openshift: true
+
+## The bundled kube-state-metrics pins uid/gid/fsGroup 65534, which is outside the namespace range.
+kube-state-metrics:
+  securityContext:
+    enabled: false
+
+## The bundled node-exporter needs hostNetwork, hostPID, hostPath volumes and host port 9100 -
+## none of which restricted-v2 permits. OpenShift's platform monitoring already runs a
+## node-exporter, so scrape that one instead.
+prometheus-node-exporter:
+  enabled: false
+nodeExporter:
+  mode: openshift
+
+## HAProxy terminates TLS on 443. restricted-v2 runs the container as a non-root uid with all
+## capabilities dropped and allowPrivilegeEscalation=false, so it cannot bind a port below 1024
+## (`cannot bind socket (Permission denied) for [0.0.0.0:443]`). Move the container port above
+## 1024. The HAProxy Service publishes this same port, so expose 443 externally through an
+## OpenShift Route or an Ingress pointing at the Service.
+haproxy:
+  containerPorts:
+    https: 8443
--- a/charts/pmm-ha/templates/_helpers.tpl
+++ b/charts/pmm-ha/templates/_helpers.tpl
@@ -203,7 +203,8 @@ stays valid across failovers.
 */}}
 {{- define "pmm.client.serverAddress" -}}
 {{- $haproxy := .Values.haproxy.fullnameOverride | default (printf "%s-haproxy" (include "pmm.fullname" .)) -}}
-{{- printf "%s.%s.svc.cluster.local:443" $haproxy .Release.Namespace -}}
+{{- $port := (.Values.haproxy.containerPorts).https | default 443 -}}
+{{- printf "%s.%s.svc.cluster.local:%v" $haproxy .Release.Namespace $port -}}
 {{- end -}}
 
 {{/*
@@ -404,4 +405,34 @@ dict with "name" and "value".
 {{- if not (regexMatch "^[A-Za-z_][A-Za-z0-9_-]*$" .value) -}}
 {{- fail (printf "%s must match ^[A-Za-z_][A-Za-z0-9_-]*$ to be usable in the ClickHouse users.d drop-in, got %q" .name .value) -}}
 {{- end -}}
+{{- end -}}
+
+{{/*
+Pod security context for the PMM Server StatefulSet.
+
+On OpenShift the namespace owns the identity: `restricted-v2` requires runAsUser to be inside
+the namespace's assigned uid-range and fsGroup inside its supplemental-group range, and rejects
+the pod outright otherwise. Emitting no securityContext at all lets OpenShift assign both.
+
+On plain Kubernetes fsGroup is load-bearing - it is what makes the PVC group-writable for the
+image's uid - so it must stay. runAsUser is not: the PMM Server image already declares
+`USER 1000`, and its entrypoint supports an arbitrary assigned uid via the NSS wrapper.
+*/}}
+{{- define "pmm.podSecurityContext" -}}
+{{- if not .Values.openshift }}
+securityContext:
+  {{- toYaml .Values.podSecurityContext | nindent 2 }}
+{{- end }}
+{{- end -}}
+
+{{/*
+Pod security context for the PMM Client StatefulSet. Same reasoning as above; the client image
+runs as uid 1002 rather than 1000.
+*/}}
+{{- define "pmm.client.podSecurityContext" -}}
+{{- if not .Values.openshift }}
+securityContext:
+  # The PMM Client image runs as this user, which has to own the volume to write to it.
+  fsGroup: {{ .Values.pmmClient.fsGroup }}
+{{- end }}
 {{- end -}}
--- a/charts/pmm-ha/values.yaml
+++ b/charts/pmm-ha/values.yaml
@@ -147,6 +147,11 @@ pmmClient:
   ## PMM Server Nodes are not offered for that.
   ##
   replicas: 3
+  ## @param pmmClient.fsGroup Group that owns the PMM Client data volume
+  ## The PMM Client image runs as this user, which has to own the volume to write to it.
+  ## Ignored when `openshift: true` - OpenShift assigns the fsGroup itself.
+  ##
+  fsGroup: 1002
   ## @param pmmClient.image.repository PMM Client image repository
   ## @param pmmClient.image.pullPolicy PMM Client image pull policy
   ## @param pmmClient.image.tag PMM Client image tag, defaults to the chart appVersion
@@ -430,6 +435,17 @@ podSecurityContext:
   runAsUser: 1000
   fsGroup: 1000
 
+## @param openshift Set to true when installing on OpenShift
+## OpenShift's `restricted-v2` SCC assigns every namespace its own uid and supplemental-group
+## ranges and rejects any pod that asks for values outside them, so the `podSecurityContext`
+## above (and `pmmClient.fsGroup`) must not be set there. With this true the chart
+## emits no pod securityContext and lets OpenShift assign the identity.
+##
+## See examples/values-openshift.yaml for the full OpenShift overlay - the bundled
+## kube-state-metrics and prometheus-node-exporter need their own settings.
+##
+openshift: false
+
 ## @param securityContext Configure Container Security Context
 ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
 ## securityContext.capabilities The capabilities to add/drop when running containers
@@ -597,6 +613,14 @@ haproxy:
   # Number of HAProxy replicas
   replicaCount: 3
 
+  ## @param haproxy.containerPorts.https Port HAProxy binds for PMM traffic
+  ## The Service publishes this same port. Keep 443 on plain Kubernetes; on OpenShift it
+  ## must be above 1024, because restricted-v2 runs the container as a non-root uid with
+  ## all capabilities dropped and so without NET_BIND_SERVICE.
+  ##
+  containerPorts:
+    https: 443
+
   ## @param haproxy.resources Resource requests and limits for the HAProxy container
   ## Every client remote_write, QAN payload and UI request crosses HAProxy, and it
   ## terminates TLS for all of them. The upstream subchart ships a request but no
@@ -634,7 +658,7 @@ haproxy:
   # so scaling up/down replicas does NOT require HAProxy restart.
   # Increment config-version only for other ConfigMap changes that require restart.
   podAnnotations:
-    pmm.percona.com/config-version: "4"
+    pmm.percona.com/config-version: "5"
 
   # HAProxy pod spread across nodes.
   #
--- a/charts/pmm-ha/templates/haproxy-configmap.yaml
+++ b/charts/pmm-ha/templates/haproxy-configmap.yaml
@@ -16,7 +16,11 @@ data:
         daemon
     
     resolvers k8s
-        nameserver dns1 kube-dns.kube-system.svc.cluster.local:53
+        # Read the cluster DNS server from the pod's own /etc/resolv.conf rather than naming
+        # a service. The vanilla-Kubernetes name kube-dns.kube-system does not exist on
+        # OpenShift (it is dns-default.openshift-dns), and HAProxy rejects an unresolvable
+        # nameserver at config-parse time, so hardcoding either name breaks the other platform.
+        parse-resolv-conf
         resolve_retries 5
         timeout resolve 2s
         timeout retry   1s
@@ -49,7 +53,12 @@ data:
     {{- end }}
     
     frontend https_front
-        bind *:443 ssl crt /etc/haproxy/certs/pmm.pem
+        # Bind port follows haproxy.containerPorts.https so it can be moved above 1024.
+        # OpenShift's restricted-v2 runs the container as a non-root uid with all
+        # capabilities dropped and allowPrivilegeEscalation=false, so NET_BIND_SERVICE is
+        # unavailable and binding 443 fails with EACCES. The Service port follows the same
+        # value, so expose 443 externally through an Ingress or an OpenShift Route.
+        bind *:{{ (.Values.haproxy.containerPorts).https | default 443 }} ssl crt /etc/haproxy/certs/pmm.pem
         default_backend https_back
 
     backend https_back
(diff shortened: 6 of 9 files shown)


charts/pmm-ha/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: pmm-ha
 description: A Helm chart for Percona Monitoring and Management (PMM)
 type: application
-version: 1.6.2
+version: 1.6.3
```
ademidoff, line 5: Convention point, not a defect: the PMM-owned charts do not get a `version:` bump in a feature PR - changes accumulate on the branch and the bump happens once in a release PR. CI does not need it either. `pmm-ha-pr-checks.yaml` runs `ct lint --check-version-increment=true` with `target-branch: main`, and `pmm-ha` on `main` is still `1.6.1`, so `1.6.2` already satisfies the check.

It does cost something concrete: #936 bumps this same line to `1.7.0` on the same base branch, so whichever lands second gets a conflict on a line that did not need to change.

```suggestion
version: 1.6.2
```

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -405,3 +406,33 @@ dict with "name" and "value".
...
+On plain Kubernetes fsGroup is load-bearing - it is what makes the PVC group-writable for the
+image's uid - so it must stay. runAsUser is not: the PMM Server image already declares
+`USER 1000`, and its entrypoint supports an arbitrary assigned uid via the NSS wrapper.
+*/}}
+{{- define "pmm.podSecurityContext" -}}
+{{- if not .Values.openshift }}
+securityContext:
+  {{- toYaml .Values.podSecurityContext | nindent 2 }}
+{{- end }}
+{{- end -}}
```
ademidoff, line 426: `openshift: true` discards the whole `podSecurityContext` map, not just the identity fields. `restricted-v2` only objects to `runAsUser` / `runAsGroup` / `fsGroup`; `supplementalGroups`, `seccompProfile` and `fsGroupChangePolicy` are all permitted, and `fsGroupChangePolicy: OnRootMismatch` in particular is exactly what someone reaches for on a large `/srv`. As written those are dropped silently, with the README saying only "Ignored when `openshift` is `true`".

Omitting just the three identity keys keeps the rest working, and it also fixes something small: because the `{{- if }}` keeps its trailing newline, `nindent 6` in the caller emits a whitespace-only line ahead of `securityContext:` in both StatefulSets, so the default render is not quite byte-identical to `PMM-HA-GA` after all:

```diff
     spec:
       serviceAccountName: pmm-service-account
+      
       securityContext:
         fsGroup: 1000
```

I rendered the replacement below against `ci/ci-values.yaml` and diffed it against `PMM-HA-GA`: no securityContext and no whitespace differences at all. Under the overlay it emits `securityContext: {}`, which admission treats the same as omitting the field.

```suggestion
{{- define "pmm.podSecurityContext" -}}
{{- $ctx := .Values.podSecurityContext | default dict -}}
{{- if .Values.openshift -}}
{{- $ctx = omit $ctx "runAsUser" "runAsGroup" "fsGroup" -}}
{{- end -}}
securityContext:
  {{- toYaml $ctx | nindent 2 }}
{{- end -}}
```

`pmm.client.podSecurityContext` just below has the same stray-blank-line behaviour; `{{- if not .Values.openshift -}}` plus `{{- end -}}` fixes it there.

charts/pmm-ha/values.yaml
```diff
@@ -430,6 +435,17 @@ podSecurityContext:
...
+## @param openshift Set to true when installing on OpenShift
+## OpenShift's `restricted-v2` SCC assigns every namespace its own uid and supplemental-group
+## ranges and rejects any pod that asks for values outside them, so the `podSecurityContext`
+## above (and `pmmClient.fsGroup`) must not be set there. With this true the chart
+## emits no pod securityContext and lets OpenShift assign the identity.
+##
+## See examples/values-openshift.yaml for the full OpenShift overlay - the bundled
+## kube-state-metrics and prometheus-node-exporter need their own settings.
+##
+openshift: false
```
ademidoff, line 447: The name promises more than the flag delivers. With `--set openshift=true` and nothing else, the render still contains:

- `kube-state-metrics` Deployment: `runAsUser: 65534`, `runAsGroup: 65534`, `fsGroup: 65534`
- `prometheus-node-exporter` DaemonSet: `hostNetwork: true`, `hostPID: true`, three `hostPath` volumes, `runAsUser: 65534`

That is rows 3 and 4 of your own table - the same "Helm reports deployed, zero pods" failure this PR exists to remove, with two of the six causes still live. The comment does point at `examples/values-openshift.yaml`, but a top-level value named `openshift` will get set on its own by someone who never opens the examples directory.

The chart already refuses to render an inconsistent OpenShift configuration: `pmm.nodeExporter.validate` fails when `nodeExporter.mode=openshift` and the bundled DaemonSet is still enabled, and `pmm-ha-pr-checks.yaml` asserts on that exact message. Extending that helper to fail on `openshift: true` with `nodeExporter.mode != "openshift"` (and, if you want to go further, with the kube-state-metrics securityContext still enabled) would make this flag either complete or loud. Fail-fast is the precedent the chart already set for precisely this class of mistake.

ademidoff (commented):
Rendered `PMM-HA-GA` and this head with `ci/ci-values.yaml`, then again with `examples/values-openshift.yaml`, and diffed all three; also rendered `--set openshift=true` on its own. The default-values path is clean (the only difference is one whitespace-only line), and both claims in the new helper comment check out against the pmm repo - `USER 1000` in `build/docker/server/Dockerfile.el9`, arbitrary-uid handling in `build/docker/server/entrypoint.sh`.

Start with the `pg-db.pmm.serverHost` comment - it looks like a fourth consumer of the HAProxy port that the overlay does not move.

charts/pmm-ha/values.yaml
```diff
@@ -430,6 +435,17 @@ podSecurityContext:
   runAsUser: 1000
   fsGroup: 1000
```
ademidoff, line 436: The comment two lines above now says something different from the flag you are adding below it:

> Note: we intentionally don't use the `runAsGroup` to make it compatible with OpenShift.
> However, `runAsUser` needs to be set to nil for OpenShift to be able to assign a random user ID.

That reads as "nil `runAsUser` is sufficient, `fsGroup: 1000` is fine on OpenShift" - which is what this PR demonstrates is wrong, since `restricted-v2`'s `fsGroup` strategy is `MustRunAs` against the namespace's supplemental-group range and rejects 1000 as well. Worth rewriting or deleting it here, so `values.yaml` does not carry two different prescriptions for the same platform.

charts/pmm-ha/examples/values-openshift.yaml
```diff
@@ -0,0 +1,33 @@
...
+  mode: openshift
+
+## HAProxy terminates TLS on 443. restricted-v2 runs the container as a non-root uid with all
+## capabilities dropped and allowPrivilegeEscalation=false, so it cannot bind a port below 1024
+## (`cannot bind socket (Permission denied) for [0.0.0.0:443]`). Move the container port above
+## 1024. The HAProxy Service publishes this same port, so expose 443 externally through an
+## OpenShift Route or an Ingress pointing at the Service.
+haproxy:
+  containerPorts:
+    https: 8443
```
ademidoff, line 33: Moving the HAProxy port leaves one consumer behind. `pg-db.pmm.serverHost` is the bare hostname `pmm-ha-haproxy` (`charts/pmm-ha/values.yaml:869`), and pmm-agent appends 443 to an address with no port - `agent/config/config.go:313` in the pmm repo, `cfg.Server.Address = net.JoinHostPort(host, "443")`. Under this overlay the HAProxy Service publishes 80 / 8443 / 1024 and nothing on 443, so the PGO PMM sidecar in every PostgreSQL pod is dialling a closed port.

You moved the other two consumers (`pmm.client.serverAddress` and `pg-pmm-token-job`'s `PMM_PORT`); this is the third. The pg-db chart exposes no port field - `serverHost` is the only knob - but the CRD in this repo types it as a plain `string` with no pattern (`charts/pg-operator/crds/crd.yaml:10846`) and pmm-agent's `SplitHostPort` accepts `host:port`, so spelling the port out should work. I rendered it: the CR comes out as `serverHost: pmm-ha-haproxy:8443`.

I could not verify the running end of this without a cluster, so a question rather than an assertion: on the ROSA run, did the PostgreSQL cluster actually appear in PMM's inventory, or only reach Running? (The `pmm-client-register` 401 in your "not in this PR" list would mask it either way.)

```suggestion
    https: 8443

## The PGO PMM sidecar in the bundled PostgreSQL cluster reaches PMM through HAProxy too,
## and pmm-agent appends :443 to an address with no port, so this one needs the port
## spelled out.
pg-db:
  pmm:
    serverHost: pmm-ha-haproxy:8443
```

charts/pmm-ha/examples/values-openshift.yaml
```diff
@@ -0,0 +1,33 @@
...
+##
+##   helm install pmm-ha percona/pmm-ha -n <namespace> -f examples/values-openshift.yaml
+##
+## OpenShift's `restricted-v2` SCC gives every namespace its own uid range and supplemental-group
+## range and rejects any pod requesting values outside them. Each setting below clears one such
+## rejection; without them the StatefulSets and Deployments are admitted-then-refused and Helm
+## still reports STATUS: deployed while no pods exist.
+
+## PMM Server and PMM Client: emit no pod securityContext, so OpenShift assigns uid and fsGroup.
+openshift: true
```
ademidoff, line 11: Nothing in CI renders this file, and nothing exercises `openshift: true`. `ct install` runs in kind, which has no SCCs, and `ct lint` only yamllints `Chart.yaml` and `values.yaml` - so the overlay and the flag have zero coverage and the first sign of drift will be a user's failed install.

The structural half of that is unavoidable (there is no OpenShift in CI), but the cheap half is right there: `pmm-ha-pr-checks.yaml` already has an offline render step ("Check node-exporter mode validation") and a `--set nodeExporter.mode=openshift` render assertion. One more `helm template ... -f charts/pmm-ha/examples/values-openshift.yaml` asserting that no pod spec keeps `runAsUser:` / `fsGroup:` and that `bind *:8443` reaches the ConfigMap would pin every mechanical part of this PR for about four lines of YAML.

charts/pmm-ha/README.md
```diff
@@ -656,6 +658,21 @@ victoriaMetrics:
...
+`NET_BIND_SERVICE`. Install with the bundled overlay, which covers all of it:
+
+```bash
+helm install pmm-ha percona/pmm-ha -n pmm -f examples/values-openshift.yaml
+```
+
+It sets `openshift: true` (PMM Server and PMM Client let the cluster assign uid and fsGroup),
+disables the bundled node-exporter in favour of OpenShift's, turns off the kube-state-metrics
+securityContext, and moves the HAProxy port to 8443. Publish 443 externally with a Route or
+Ingress pointing at the HAProxy Service.
```
ademidoff, line 674: This section is right, but it now contradicts four lines earlier in the same file that still hardcode HAProxy on 443:

- L287 - `| pmm-ha-haproxy | ... | 443 (HTTPS) |`
- L298 and L301 - `pmm-admin config --server-url=https://...@pmm-ha-haproxy:443 ...`
- L735 - "Connect via `https://<EXTERNAL-IP>:443`"

All four are wrong for an install that used this overlay. A parenthetical - "or `haproxy.containerPorts.https`, e.g. 8443 on OpenShift" - on each would settle it.

Smaller second point: "an Ingress pointing at the HAProxy Service" is not what the chart's own `ingress.enabled: true` produces. `templates/ingress.yaml` points at `.Values.service.name` (`monitoring-service`, 8443), bypassing HAProxy entirely. Worth saying the Route or Ingress has to be hand-written, or a reader will flip the chart's switch and end up pinned to one PMM pod with no leader routing.

charts/pmm-ha/values.yaml
```diff
@@ -147,6 +147,11 @@ pmmClient:
   ## PMM Server Nodes are not offered for that.
   ##
   replicas: 3
+  ## @param pmmClient.fsGroup Group that owns the PMM Client data volume
+  ## The PMM Client image runs as this user, which has to own the volume to write to it.
+  ## Ignored when `openshift: true` - OpenShift assigns the fsGroup itself.
+  ##
+  fsGroup: 1002
```
ademidoff, line 154: `CLAUDE.md`: "`values.yaml` uses `## @section` / `## @param` annotations, and each chart's `README.md` carries a hand-maintained parameter table generated from them. Adding or renaming a value means updating both." `openshift` and `haproxy.containerPorts.https` both made it into the table; this one did not - and it has nowhere to go, because the table has no `pmmClient.*` rows at all. Pre-existing gap, but this is the first `pmmClient` knob a user has a concrete reason to change, so it is a reasonable moment to add the row.

charts/pmm-ha/examples/values-openshift.yaml
```diff
@@ -0,0 +1,33 @@
...
+  enabled: false
+nodeExporter:
+  mode: openshift
+
+## HAProxy terminates TLS on 443. restricted-v2 runs the container as a non-root uid with all
+## capabilities dropped and allowPrivilegeEscalation=false, so it cannot bind a port below 1024
+## (`cannot bind socket (Permission denied) for [0.0.0.0:443]`). Move the container port above
+## 1024. The HAProxy Service publishes this same port, so expose 443 externally through an
+## OpenShift Route or an Ingress pointing at the Service.
+haproxy:
```
JiriCtvrtka, line 31: pg-db.pmm.serverHost has no port while the overlay moves HAProxy to 8443. The PG operator copies that value verbatim into PMM_AGENT_SERVER_ADDRESS, pmm-agent appends :443, and the sidecar can never connect. It fails silently — the token job was updated and succeeds, so the install looks healthy while the bundled PostgreSQL goes unmonitored. This is on the exact path the PR exists to enable, so it should not land as-is. One-line fix in the overlay: serverHost: pmm-ha-haproxy.<ns>.svc.cluster.local:8443.

charts/pmm-ha/Chart.yaml
theTibi: Reverted to 1.6.2 in bab85b1.

charts/pmm-ha/templates/_helpers.tpl
theTibi: Fixed in bab85b1 with your `omit` version. Confirmed the blank line: rendering the helper in isolation gave a whitespace-only line ahead of `securityContext:`, so the "byte-identical" claim in the description was wrong. After the change the default render diffs clean against PMM-HA-GA.

`pmm.client.podSecurityContext` builds its map inline rather than from values, so `omit` does not apply there; it now branches on the flag and emits `securityContext: {}` on OpenShift.

Verified the keys you named survive: with `--set podSecurityContext.fsGroupChangePolicy=OnRootMismatch --set podSecurityContext.seccompProfile.type=RuntimeDefault` under the overlay, both render and only the identity keys are dropped.

charts/pmm-ha/values.yaml
theTibi: Added `pmm.openshift.validate` in bab85b1, following the `pmm.nodeExporter.validate` precedent. `openshift: true` now fails to render unless `nodeExporter.mode=openshift` and `kube-state-metrics.securityContext.enabled=false`, skipping the second check when kube-state-metrics is disabled outright. Both messages point at the overlay. CI asserts on both.

charts/pmm-ha/values.yaml
theTibi: Rewritten in bab85b1. The note now says `restricted-v2` rejects `fsGroup: 1000` as well, its fsGroup strategy being MustRunAs against the namespace range.

charts/pmm-ha/examples/values-openshift.yaml
theTibi: Answering your question: the cluster only reached Running. Verified on a plain Kubernetes cluster with `haproxy.containerPorts.https: 8443`, which exercises the same path without needing OpenShift.

Before, with the bare hostname:

```
PMM_AGENT_SERVER_ADDRESS = pmm-ha-haproxy
Updating PMM Server address from "pmm-ha-haproxy" to "pmm-ha-haproxy:443".
Registering pmm-agent on PMM Server...
Failed to register pmm-agent on PMM Server: Post "https://pmm-ha-haproxy:443/v1/management/nodes": dial tcp 10.100.29.0:443: i/o timeout.
Restarting `pmm-agent setup` in 5 seconds because PMM_AGENT_SIDECAR is enabled...
```

That loops forever. The Service published 80/8443/1024 and the PG pod stayed 5/5 Running the whole time, with `/v1/inventory/services` returning `"postgresql": []`.

After `serverHost: pmm-ha-haproxy:8443`:

```
Registering pmm-agent on PMM Server...
Registered.
Creating gRPC client to https://service_token:***@pmm-ha-haproxy:8443/ ...
```

No address rewrite at all, the service is in the inventory and `postgres_exporter` is AGENT_STATUS_RUNNING. Fixed in bab85b1 with your value - the short name resolves since pg-db lands in the release namespace.

The `pmm-client-register` 401 did not mask it here: the token job completed and wrote `pg-pmm-secret`, so this run isolates the serverHost defect on its own.

charts/pmm-ha/examples/values-openshift.yaml
theTibi: Added in bab85b1. It had to go in the `--dry-run=server` job rather than the offline one, since a successful render hits the `pmm-secret` lookup. It asserts no identity keys in either StatefulSet under the overlay, that both keep theirs without it, `bind *:8443`, all four consumers of the port, and that neither subchart renders. Two fail-fast cases for the flag went into the existing offline step.

charts/pmm-ha/README.md
theTibi: Fixed in bab85b1 - all four now name `haproxy.containerPorts.https`.

Your second point was the more useful one: the section now says to use an OpenShift Route against the HAProxy Service and warns that `ingress.enabled` targets `monitoring-service` and bypasses HAProxy. Same correction applied to the overlay comment and `haproxy-configmap.yaml`.

charts/pmm-ha/values.yaml
theTibi: Added in bab85b1. Since the table had no `pmmClient.*` rows at all, I added the section covering all 12 params rather than just this one.

charts/pmm-ha/examples/values-openshift.yaml
theTibi: Answering your question: the cluster only reached Running. Verified on a plain Kubernetes cluster with `haproxy.containerPorts.https: 8443`, which exercises the same path without needing OpenShift.

Before, with the bare hostname:

```
PMM_AGENT_SERVER_ADDRESS = pmm-ha-haproxy
Updating PMM Server address from "pmm-ha-haproxy" to "pmm-ha-haproxy:443".
Registering pmm-agent on PMM Server...
Failed to register pmm-agent on PMM Server: Post "https://pmm-ha-haproxy:443/v1/management/nodes": dial tcp 10.100.29.0:443: i/o timeout.
Restarting `pmm-agent setup` in 5 seconds because PMM_AGENT_SIDECAR is enabled...
```

That loops forever. The Service published 80/8443/1024 and the PG pod stayed 5/5 Running the whole time, with `/v1/inventory/services` returning `"postgresql": []`.

After `serverHost: pmm-ha-haproxy:8443`:

```
Registering pmm-agent on PMM Server...
Registered.
Creating gRPC client to https://service_token:***@pmm-ha-haproxy:8443/ ...
```

No address rewrite at all, the service is in the inventory and `postgres_exporter` is AGENT_STATUS_RUNNING. Fixed in bab85b1 with your value - the short name resolves since pg-db lands in the release namespace.

The `pmm-client-register` 401 did not mask it here: the token job completed and wrote `pg-pmm-secret`, so this run isolates the serverHost defect on its own.

charts/pmm-ha/templates/_helpers.tpl
```diff
@@ -303,6 +304,29 @@ Called from statefulset.yaml, which always renders.
...
+Fail-fast validation for the `openshift` flag.
+
+It only governs the PMM Server and PMM Client pod securityContexts. The bundled
+kube-state-metrics and prometheus-node-exporter carry their own `restricted-v2` violations, and
+left at their defaults they reproduce the very failure the flag exists to remove: Helm reports
+`STATUS: deployed` while those workloads are rejected at admission and produce zero pods. So the
+flag requires the rest of the overlay rather than silently delivering half of it.
+Called from statefulset.yaml, which always renders.
+*/}}
+{{- define "pmm.openshift.validate" -}}
```
JiriCtvrtka, line 384: Blocker: pmm.openshift.validate fails the install if node-exporter or kube-state-metrics are left at their non-OpenShift defaults, but says nothing about haproxy.containerPorts.https. So the documented happy path (openshift=true + the two required overrides, no overlay file) renders bind *:443, Helm reports success, and HAProxy — the only ingress into the cluster — crash-loops under restricted-v2. That's a green install with a dead product, which is worse than a hard failure. Fix is one more clause in the same validator (or defaulting the port when openshift=true).

theTibi: Confirmed and fixed in 5cfeee2a.

Added the clause, keyed on `lt 1024` rather than `eq 443` — 80 and 1023 rendered just as happily. Went with the validator over defaulting the port: silently moving it would change the Service port under an existing Route and re-break `pg-db.pmm.serverHost`.

That second one turned out to be the same bug without the OpenShift part, so there is now a `pmm.haproxy.validate` too. The operator copies `serverHost` verbatim and pmm-agent appends :443 when the address has no port, so moving the port on plain Kubernetes leaves the PG sidecar dialling a closed port with every pod Running and the inventory empty. It only fires when `serverHost` points at this chart's own HAProxy.

Verified on ROSA HCP, OpenShift 4.21.31. Before the fix, `openshift=true` plus the two overrides the validator named installs green and HAProxy crash-loops with `cannot bind socket (Permission denied) for [0.0.0.0:443]`; the same namespace at 8443 runs 3/3. A full install with the overlay reaches green, PG sidecars reporting through `pmm-ha-haproxy:8443`. Plain-Kubernetes default render is byte-identical. Six cases pinned in CI.

Also fixed the README paragraph that said the flag needs two things.

PR #951, opened 2026-09-04
PMM-15445 Settle concurrent pmm-token-init runs on the secret

theTibi:
Rebuilt on current `PMM-HA-GA`. #941 landed the hashed Job name, which was this ticket's
original scope, so that part is dropped here. What remains is the case #941 does not cover.

## Problem

More than one `pmm-token-init` pod can be running at once. `Client.Update` applies the whole
target list before pruning `original.Difference(target)`, with `DeletePropagationBackground`,
so the new Job is scheduled while the old one is still running. A prune that fails is logged
and skipped, and a Job created by a revision that then failed is not in the next upgrade's
diff at all (that diffs against the last *deployed* revision), so it is never pruned.

Two pods that get that far both mint a token and both write the secret with
`kubectl apply`. Last writer wins and the loser's Admin token stays valid in Grafana,
referenced by nothing.

Separately, the README added in #941 tells people to select the Job on
`app.kubernetes.io/component: pmm-token-init`. Nothing emits that label - the Job carries
`component: pmm-server` and the pods carry no chart labels at all - so the documented
command returns nothing.

## Fix

- Resolve the race on the secret. `kubectl create` is the only atomic step available, so it
  decides the winner rather than a read followed by a write, and a run that loses revokes
  the token it had minted. Only that run's own token is deleted; revoking every token on the
  service account would pull the rug out from under PostgreSQL pods that already have the
  live one mounted.
- `app.kubernetes.io/component: pmm-token-init` on the Job and on its pods, derived from
  `pmm.labels` by substitution so the two cannot drift. Makes the README's command work.
- CI asserts that the name hash *tracks the pod spec*: it moves for a value that feeds the
  pod spec and holds still for one that does not. Shape alone proves nothing - hashing a
  constant passes that. The existing job only does a fresh `ct install`, which never
  exercises an upgrade.

### Why the comparison is against `STALE_TOKEN` and not just "is the secret non-empty"

`kubectl create` failing means the secret exists, which covers three different situations.
Reading the token back tells them apart:

| Token now in the secret | Meaning | Action |
|---|---|---|
| different from `STALE_TOKEN` | another run installed one behind this one | defer to it, revoke this run's token |
| equal to `STALE_TOKEN` | the dead token this run is replacing (#941's re-mint path) | overwrite |
| empty | pre-existing secret carrying no token | overwrite |

Without the comparison the second row looks like a lost race, so the re-mint path revokes the
good token it has just minted and leaves the dead one in the secret - permanently, since every
retry repeats it. Reproduced live, see the table below.

## Verified

Live on a 1.36 EKS cluster against a running PMM HA, using the real templates in a scratch
namespace. `PMM-HA-GA` is the same test on unmodified upstream.

| Test | `PMM-HA-GA` | This branch |
|---|---|---|
| Fresh install | token minted, `HTTP 200` | token minted, `HTTP 200` |
| README's documented label selector | **0 jobs matched**, `No resources found` | job and logs both returned |
| `helm upgrade` onto this branch | n/a | deployed in 4.3s, Job renamed, old pruned, **no duplicate token** |
| No-op `helm upgrade` | n/a | same Job uid, same pod - no recreate, no re-pull |
| Dead token in the secret | re-mints, `HTTP 200` | re-mints, `HTTP 200` |
| Dead token, without the `STALE_TOKEN` comparison | n/a | **leaves the dead token**, revokes the good one |
| Two pods racing one secret | **+2 tokens**, one orphaned | **+1 token**, loser revoked cleanly |

The revoke uses the service account UID in the path, matching the create call three lines
above it; confirmed against Grafana, `HTTP 200`, no `WARNING` line in the Job log.

The CI step was also run with `--dry-run=server` against that cluster, not just offline:
all assertions pass, `base job name = pmm-2-pmm-token-init-6c52f224`.

## How to test

```sh
NS=<namespace>; REL=<release>
SEL="app.kubernetes.io/instance=$REL,app.kubernetes.io/component=pmm-token-init"

# 1. The README's command works (returns nothing on PMM-HA-GA)
kubectl get job -n "$NS" -l "$SEL"
kubectl logs  -n "$NS" -l "$SEL"

# 2. A dead token is replaced, not preserved
kubectl patch secret "$REL-pg-db-pmm-secret" -n "$NS" --type merge \
  -p "{\"data\":{\"PMM_SERVER_TOKEN\":\"$(printf 'glsa_dead' | base64)\"}}"
# re-run the Job, then confirm the secret holds a token that authenticates:
TOK=$(kubectl get secret "$REL-pg-db-pmm-secret" -n "$NS" -o jsonpath='{.data.PMM_SERVER_TOKEN}' | base64 -d)
curl -sk -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer $TOK" \
  "https://pmm-ha-haproxy.$NS.svc.cluster.local/graph/api/user"      # expect 200

# 3. Racing pods leave exactly one new token. Count before and after:
curl -sk -u "admin:$PMM_ADMIN_PASSWORD" \
  "https://pmm-ha-haproxy.$NS.svc.cluster.local/graph/api/serviceaccounts/<sa-uid>/tokens"
```

No `Chart.yaml` bump, matching #941 and #944.

The red `lint-test` from **Lint and Test Charts** is pre-existing and unrelated: `test.yaml`
calls `ct list-changed` without `--target-branch`, so a PR based on `PMM-HA-GA` is diffed
against `main` and picks up `charts/pg-db` from #918. #970 fixes this in
`pmm-ha-pr-checks.yaml`; `test.yaml` still has it in four places.


charts/pmm-ha/templates/pg-pmm-token-job.yaml
```diff
@@ -186,4 +182,53 @@ spec:
...
+  name: {{ .Release.Name }}-pmm-token-manager
+subjects:
+- kind: ServiceAccount
+  name: {{ .Release.Name }}-pmm-token-generator
+  namespace: {{ .Release.Namespace }}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ printf "%s-pmm-token-init" .Release.Name | trunc 54 | trimSuffix "-" }}-{{ include "pmm.pgTokenJobPodTemplate" . | sha256sum | trunc 8 }}
```
JiriCtvrtka, line 226: Hashing the pod spec into the name fixes the immutability failure, but it also turns "upgrade fails" into "Job gets replaced mid-run", and the script isn't idempotent about token creation yet.

The early exit at L79–82 only fires once the secret already holds a non-empty PMM_SERVER_TOKEN. Before that, every run unconditionally mints a new token — TOKEN_NAME="pg-monitoring-token-$(date +%s)" (L155) → POST /serviceaccounts/${SA_UID}/tokens (L160) — against a service account created with "role":"Admin" (L121), and only writes the secret at the very end (L176–179).

So between L160 and L179 the token is live in PMM with nothing referencing it. With the unbounded readiness loop at L92, that window is realistically long: an upgrade that touches the pod spec while PMM isn't ready yet replaces the Job, and the old pod dies holding an Admin token that nothing will ever revoke. ttlSecondsAfterFinished reaps the Job, not the token. If Helm's apply-then-prune leaves both Jobs alive briefly, both mint a token and only one reaches the secret.

Pre-PR this couldn't happen: the upgrade failed and the Job stayed put.

Cheapest fix is to make token creation idempotent — the script already resolves SA_UID, so it can list and delete existing tokens before minting:


EXISTING=$(curl -sk -u "admin:${PMM_ADMIN_PASSWORD}" \
  "${PMM_URL}/graph/api/serviceaccounts/${SA_UID}/tokens" \
  | grep -o '"id":[0-9]*' | cut -d: -f2)
for id in $EXISTING; do
  curl -sk -X DELETE -u "admin:${PMM_ADMIN_PASSWORD}" \
    "${PMM_URL}/graph/api/serviceaccounts/${SA_UID}/tokens/${id}" >/dev/null
done
Not a Helm hook — your reasoning about the unbounded wait blocking the release is right. Happy to be wrong about Helm's ordering if you've watched an upgrade do this.

.github/workflows/pmm-ha-pr-checks.yaml
```diff
@@ -158,6 +158,13 @@ jobs:
             | grep -qF "regex: 'pmm\\.ha'" \
             || { echo "::error::Release.Name is not regex-escaped in vmagent scrape configs"; exit 1; }
 
+          # The token-init Job name must carry a hash of its pod spec. A fixed name makes any
+          # upgrade that touches that pod spec fail on spec.template immutability (PMM-15445).
+          helm template pmm-2 charts/pmm-ha --namespace "$NS" --dry-run=server \
+            -s templates/pg-pmm-token-job.yaml \
+            | grep -qE '^  name: pmm-2-pmm-token-init-[0-9a-f]{8}$' \
+            || { echo "::error::pmm-token-init Job name lost its pod-spec hash"; exit 1; }
```
ademidoff, line 166: The assertion checks the *shape* of the name, not the property the name has to have. I patched a scratch copy of the chart to hash a literal instead of the pod template, then ran this grep verbatim against it:

```
$ # name: {{ printf ... }}-{{ "constant" | sha256sum | trunc 8 }}
$ helm template pmm-2 charts/pmm-ha -s templates/pg-pmm-token-job.yaml \
    | grep -qE '^  name: pmm-2-pmm-token-init-[0-9a-f]{8}$' && echo PASSES
PASSES
```

with the suffix frozen across a pod-spec change that must move it:

```
constant hash, pg-db.pmm.secret=pg-pmm-secret -> pmm-2-pmm-token-init-96f6549c
constant hash, pg-db.pmm.secret=hash-probe    -> pmm-2-pmm-token-init-96f6549c
this branch,   pg-db.pmm.secret=pg-pmm-secret -> pmm-2-pmm-token-init-dab73e63
this branch,   pg-db.pmm.secret=hash-probe    -> pmm-2-pmm-token-init-e26ef8b9
```

The two properties that make the fix work — the suffix moves when the pod spec moves, and holds still when it does not — are exactly the ones your description verifies by hand and CI does not. They are two more renders. Worth saying plainly too: nothing in this workflow runs `helm upgrade`, so the PMM-15445 failure itself stays uncovered and these renders are a proxy for it.

Also, this re-renders the whole chart through the API a fourth time to read one line that `$out` from L126 already contains (same release name, same values).

```suggestion
          # The token-init Job name must carry a hash of its pod spec. A fixed name makes any
          # upgrade that touches that pod spec fail on spec.template immutability (PMM-15445).
          grep -qE '^  name: pmm-2-pmm-token-init-[0-9a-f]{8}$' <<<"$out" \
            || { echo "::error::pmm-token-init Job name lost its pod-spec hash"; exit 1; }

          # ...and the hash has to track the pod spec: a value that feeds it must move the
          # suffix, and a value that does not must leave it alone.
          job_name() {
            helm template pmm-2 charts/pmm-ha --namespace "$NS" --dry-run=server \
              -s templates/pg-pmm-token-job.yaml "$@" \
              | awk '/^kind: Job/{f=1} f && /^  name:/{print $2; exit}'
          }
          base=$(job_name)
          [[ "$(job_name --set pg-db.pmm.secret=hash-probe)" != "$base" ]] \
            || { echo "::error::Job name did not move when the pod spec changed"; exit 1; }
          [[ "$(job_name --set replicaCount=5)" == "$base" ]] \
            || { echo "::error::Job name churns on a value outside the pod spec"; exit 1; }
```

I ran all three assertions against this branch's chart and they pass (`dab73e63` / `e26ef8b9` / `dab73e63`).

charts/pmm-ha/templates/pg-pmm-token-job.yaml
```diff
@@ -1,51 +1,47 @@
...
+{{/*
+The pod template of the token-init Job, kept in a named template so the Job's name can carry a
+hash of it.
+
+A Job's spec.template is immutable. With a fixed name, any chart change that moves this pod spec
+- a new SECRET_NAME, a different port, an edited script - makes `helm upgrade` fail with
+"spec.template: Invalid value: ...: field is immutable" instead of updating the Job (PMM-15445).
+Hashing the template into the name means a changed spec produces a NEW Job, and Helm deletes the
+old one because it is no longer in the manifest; an unchanged spec keeps the same name, so a
+routine upgrade does not recreate the Job or re-pull its image.
```
ademidoff, line 31: "Helm deletes the old one because it is no longer in the manifest" is true, but not as a handover — it creates the new Job first and prunes the old one afterwards, so the two overlap. `pkg/kube/client.go`, `update()`, v3.18.0:

```go
err := target.Visit(func(info *resource.Info, err error) error {
        ...
        // Since the resource does not exist, create it.
        if err := createResource(info); err != nil {          // <- new Job created here
                return errors.Wrap(err, "failed to create resource")
        }
        ...
})
switch {
case err != nil:
        return res, err
case len(updateErrors) != 0:
        return res, errors.New(strings.Join(updateErrors, " && "))
}

for _, info := range original.Difference(target) {            // <- old Job pruned only here
        ...
        if err := deleteResource(info, metav1.DeletePropagationBackground); err != nil {
                c.Log("Failed to delete %q, err: %s", info.ObjectName(), err)
                continue
        }
}
```

Three things follow, which answer @JiriCtvrtka's "happy to be wrong about Helm's ordering" without needing to watch an upgrade:

- the entire `target` list is created and updated before any pruning starts, so the new Job is scheduled while the old one is still running;
- the prune uses `metav1.DeletePropagationBackground`, so the old Job object disappears at once but its pod is reaped asynchronously by the GC — the old pod can outlive the Job object and keep running mid-`curl`. Helm does not wait for it either way (`u.Wait` in `pkg/action/upgrade.go` is applied at L446, after `KubeClient.Update` at L429, and only waits on the target resources);
- a failed prune is logged and `continue`d, not returned as an error, so an upgrade can report success with the old Job still in place.

The comment is the natural place to record when the old Job goes away, since that timing is the whole basis of the idempotency problem in the other thread. Read from the pinned v3.18.0 source; not observed against a cluster.

charts/pmm-ha/templates/pg-pmm-token-job.yaml
ademidoff, line 226: `trunc 54` is applied to the joined string, so what gets eaten is `-pmm-token-init`, not the release name. Rendered from this branch at increasing release-name lengths:

```
release len  job name                                                job len
39           aaaa...aaa-pmm-token-init-7fda0e20                      63
40           aaaa...aaaa-pmm-token-ini-6d248f14                      63
50           aaaa...aaaaaaaaaaaaa-pmm-e1304791                       63
52           aaaa...aaaaaaaaaaaaaaa-p-781422c2                       63
53           aaaa...aaaaaaaaaaaaaaaaa-4b3fef13                       62
```

The 63-character cap holds everywhere — `helm` itself rejects anything longer than 53 for a release name, so 63 is never exceeded — and the names stay unique, so nothing breaks. But past 39 characters the name stops saying what it is, and at the release-name limit it is just `<release>-<hash>` with no marker at all. That also makes `<release>-pmm-token-init-<hash>` in `README.md` and the CI grep descriptions of the short-name case only.

Truncating the release name instead of the joined string keeps the marker at every length, under the same ceiling:

```suggestion
  name: {{ printf "%s-pmm-token-init" (.Release.Name | trunc 39 | trimSuffix "-") }}-{{ include "pmm.pgTokenJobPodTemplate" . | sha256sum | trunc 8 }}
```

I rendered that at lengths 2, 5, 38, 39, 40, 50, 52 and 53: byte-identical to this branch for every release name up to 39 characters (so no rename for existing releases, `t1-pmm-token-init-c2edcb0c` and `pmm-2-pmm-token-init-dab73e63` unchanged), `<release>-pmm-token-init-<hash>` for all the rest, longest output 63.

charts/pmm-ha/templates/pg-pmm-token-job.yaml
```diff
@@ -186,4 +182,53 @@ spec:
...
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ printf "%s-pmm-token-init" .Release.Name | trunc 54 | trimSuffix "-" }}-{{ include "pmm.pgTokenJobPodTemplate" . | sha256sum | trunc 8 }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "pmm.pgTokenJob.labels" . | nindent 4 }}
+spec:
+  ttlSecondsAfterFinished: 86400  # Clean up after 24 hours
```
ademidoff, line 231: `ttlSecondsAfterFinished` only reaps *finished* Jobs (Complete or Failed), and the readiness loop at L92 is unbounded by design — so a Job that Helm stops tracking while PMM is still coming up never finishes and never expires. That did not matter before this PR: with one fixed name there was only ever one `<release>-pmm-token-init`.

Helm can stop tracking one. `pkg/action/upgrade.go` (v3.18.0) chooses the manifest to diff against like this:

```go
if lastRelease.Info.Status == release.StatusDeployed {
        currentRelease = lastRelease
} else {
        // finds the deployed release with the given name
        currentRelease, err = u.cfg.Releases.Deployed(name)
```

So after a *failed* upgrade the next upgrade diffs against the last **deployed** revision, not the failed one. A Job that the failed revision created under its own hash is in neither `original` nor `target` next time round, so `original.Difference(target)` never names it and it is never pruned.

Reaching that state is ordinary rather than exotic in this chart:

- `--wait` (or `--atomic`, which sets it) timing out — `KubeClient.Update` at L429 has already applied everything, including the new Job, before the wait at L446 fails the release;
- any manifest that fails to apply *after* the Job. `pkg/releaseutil/kind_sorter.go` puts unrecognised kinds last ("unknown kind is last"), which is every CR this chart renders — `VMCluster`, `VMAgent`, `VMAuth`, `ClickHouseInstallation`, `PerconaPGCluster`. A webhook rejecting one of those returns from `update()` at the `len(updateErrors) != 0` branch, i.e. before the prune loop runs at all.

It does self-heal once the secret holds a token — the L79-82 early exit lets the orphan complete, and then the TTL takes it — so the exposure is bounded by "PMM eventually becomes ready", and it overlaps with the mid-run replacement @JiriCtvrtka raised. Flagging it because it is a different mechanism (permanent, from a *failed* revision, not a brief overlap during a successful one) and because it means the TTL no longer bounds how many token Jobs can exist. Reasoned from the two Helm code paths above; not reproduced, which would need a cluster.

ademidoff (commented):
Verified by rendering base and head with `helm template` (the `lookup` in `pg-user-credentials-secrets.yaml` dropped from a scratch copy; no cluster contacted): the only non-random difference is the three lines you claim, the suffix is deterministic and moves only for `secret.name` and `pg-db.pmm.secret` — not for `replicaCount`, `image.tag`, `resources`, `extraLabels`, the namespace, or a `Chart.yaml` version bump — and the 63-character cap holds at every legal release-name length. Look at the CI assertion first: I reproduced it passing against a chart whose hash is a literal constant. The Helm ordering question left open in @JiriCtvrtka's thread is settled below from the pinned v3.18.0 source — create-then-prune, so the overlap is real.

charts/pmm-ha/templates/pg-pmm-token-job.yaml
theTibi: Confirmed and applied verbatim in b7d961c. At a 45-character release name: `…-pmm-toke-ff90efe1` → `…-pmm-token-init-6a57f9b3`.

Installed live at that length — the API server accepted the 63-character name and the Job completed. Short-name releases render unchanged.

.github/workflows/pmm-ha-pr-checks.yaml
theTibi: Right, and I reproduced it: the shape-only grep passes against a chart that hashes a constant. Took your block in b7d961c, with `base` read from `$out` rather than a separate render, so it is two extra renders instead of three.

Checked both directions:

| chart | old assertion | new assertion |
|---|---|---|
| this branch | pass | pass |
| hash of a constant | **pass** | fails — "did not move" |
| hash folds in `replicaCount` | pass | fails — "churns" |

Also ran the new block verbatim against the real chart with `--dry-run=server` on a live cluster: passes.

charts/pmm-ha/templates/pg-pmm-token-job.yaml
theTibi: Agreed — comment reworded in b7d961c. I also watched it on a cluster rather than only reading the source; polling every 200ms across the upgrade caught both Jobs present:

```
jobs=1  t1-pmm-token-init-c2edcb0c
jobs=2  t1-pmm-token-init-875c1746
jobs=1  t1-pmm-token-init-875c1746
```

The script no longer assumes it is the only token-init pod — it settles on the secret instead of on the Job.

charts/pmm-ha/templates/pg-pmm-token-job.yaml
theTibi: Correct, and it is the mechanism the fix in b7d961c is built around. Rather than trying to bound the Job, the script now assumes another token-init pod may be alive: it re-checks the secret after the readiness loop, **before** minting, so an orphan that wakes up late exits without creating a token.

Live A/B on exactly that scenario (secret populated by a winner while the pod sat in the readiness loop):

| | secret after | tokens on the Admin SA |
|---|---|---|
| before the fix | winner's token clobbered | 10 → 11 |
| after | winner's token intact | 11 → 11 |

charts/pmm-ha/templates/pg-pmm-token-job.yaml
theTibi: Real problem, fixed in b7d961c — but not with delete-all, which would revoke the live token. An orphan passes the early exit long before it reaches the mint step, so when PMM finally comes up it would delete every token on the service account, including the one already in the secret and mounted by the PostgreSQL pods.

Instead: re-check the secret once PMM is ready and before minting, then install the token with an atomic `kubectl create`. A run that loses that race deletes only the token id it minted itself.

Two corrections to the framing. The window is short — the unbounded loop runs before any token exists, so a replacement during the wait leaks nothing. And the leak is not new here: `restartPolicy: OnFailure` + `backoffLimit: 10` means any failure after the mint re-runs the script and mints again, which is the behaviour on `PMM-HA-GA` today.

Evidence, against a real PMM:

- orphan scenario before/after — secret clobbered and SA 10 → 11 tokens, versus secret intact and count unchanged;
- withdraw path, with the mint-to-write window widened to hit it deterministically: token 78 minted, racer won the secret, only token 78 deleted, count back to baseline, racer's token still in the secret.

That service account already had 10 Admin tokens on it, 8 predating the test and most with `lastUsedAt: null` — so the leak is happening in practice, mostly through the retry path rather than Job replacement.

JiriCtvrtka (approved):

charts/pmm-ha/templates/pg-pmm-token-job.yaml
```diff
@@ -222,14 +263,49 @@ embedded shell script is untouched.
...
+          # Only the token minted by THIS run is ever deleted: revoking every token on the
+          # service account would pull the rug out from under PostgreSQL pods that already have
+          # the live one mounted.
+          if kubectl create secret generic "$SECRET_NAME" \
+               --namespace="$NAMESPACE" \
+               --from-literal=PMM_SERVER_TOKEN="$TOKEN_KEY" >/dev/null 2>&1; then
+            echo "Secret '$SECRET_NAME' created with PMM_SERVER_TOKEN"
+          else
+            WINNING_TOKEN=$(current_token)
+            if [ -n "$WINNING_TOKEN" ] && [ "$WINNING_TOKEN" != "$STALE_TOKEN" ]; then
```
JiriCtvrtka: High: This guard can't distinguish "no race" from "I lost the race" on the re-mint path. STALE_TOKEN is read at line 188, before minting. When the secret already holds a dead token and two runs overlap, neither has written yet at line 289, so both read WINNING_TOKEN equal to STALE_TOKEN, the guard is false for both, and both fall through to the apply at 301. Last writer wins and the other run's Admin token stays live in PMM, referenced by nothing — the leak this PR is meant to close. The comment at 275-278 is right that equality means "the dead token I'm replacing", but it's also what a lost race looks like from here. Fresh install is genuinely fixed, since the kubectl create at 284 arbitrates atomically; upgrade/re-mint is not, and README:337 claims it is.

Suggested direction: let the write arbitrate rather than the pre-read — optimistic concurrency on resourceVersion (or a precondition patch) so exactly one run's write lands, then have every run revoke its own token unless it reads that token back out of the secret afterwards. That also collapses the fresh-install and re-mint paths into one.

theTibi: Confirmed, and reproduced on a cluster. Fixed in 69bcca07.

Two Jobs started together don't actually race — the 10s readiness poll preserves their start offset, so one finishes before the other reads. I forced it with a proxy that holds `/v1/readyz` open until both pods are waiting and then releases them together. Against a secret holding a dead token, the code on this branch leaked in **5/5** trials: both pods re-minted, both applied, neither withdrew, two Admin tokens live on the service account with the secret referencing one.

The fix is the direction you suggested — the write arbitrates, not the pre-read. The secret is installed either by the `create` (absent) or by a JSON Patch carrying a `test` on the resourceVersion the token was read at (present), so of two runs holding the same version exactly one write lands. The loser re-reads, finds a token other than the one it meant to replace, withdraws the token it minted and defers. Running out of attempts withdraws it too rather than leaking, and fails so the Job retries.

A patch rather than `replace`: `replace` is a whole-object PUT and would drop other keys, labels and the last-applied annotation. RBAC already grants `patch`, so the Role is unchanged.

Verified live: 8/8 two-pod races and 6/6 three-pod races end with exactly one live token, and it authenticates. Also covered fresh install, a secret with no `.data`, a secret carrying a foreign key (preserved), an already-valid token (early exit, no mint), and the give-up path with `patch` removed from the Role — 5 attempts, own token revoked, nothing leaked, Job fails for retry.

Two things came out of the testing:

- `TOKEN_NAME` was built at one-second resolution, and Grafana requires token names unique within the org. Two pods minting in the same second sent the same name; the second got a 400 with no key in the response and failed with `Could not create service account token`, nothing saying why. This was masking the leak — it turned 4 of my first 5 collisions into a failed Job instead. Now carries the pod name, which also records which run minted a given token.
- A rejected write now passes kubectl's own message through. An RBAC or validation failure looked identical to a lost race from here, and those don't resolve on a retry.

README:344-349 was the overclaiming paragraph (line 337 is the code fence) — rewritten to say the write settles it.

charts/pmm-ha/templates/pg-pmm-token-job.yaml
```diff
@@ -222,14 +290,105 @@ embedded shell script is untouched.
...
-          kubectl create secret generic "$SECRET_NAME" \
-            --namespace="$NAMESPACE" \
-            --from-literal=PMM_SERVER_TOKEN="$TOKEN_KEY" \
-            --dry-run=client -o yaml | kubectl apply -f -
+          # Withdraw the token minted by THIS run. Only ever this run's own token: revoking every
+          # token on the service account would pull the rug out from under PostgreSQL pods that
+          # already have the live one mounted.
+          revoke_own_token() {
+            if [ -n "$TOKEN_ID" ]; then
+              curl -sk -X DELETE \
```
4nte: Not a blocker but small and easy to fix:
curl without -f exits 0 on any HTTP response, so the || echo "WARNING..." only fires on connection errors. If the DELETE gets a 503 back (for example from HAProxy while there's no healthy leader during the same helm upgrade that caused the overlap), the token stays live in Grafana and the log says nothing.

It only matters in the rare lost-race path, and a leaked token doesn't break anything, so it's not blocking. But since the point of this path is to avoid orphaned Admin tokens, it would be nice to see when that fails. It would also mean "no WARNING line" actually shows the revoke worked.

Matching the %{http_code} pattern the script already uses for the readiness and token probes:
```
DEL_CODE=$(curl -sk -o /dev/null -w "%{http_code}" -X DELETE \
  -u "admin:${PMM_ADMIN_PASSWORD}" \
  "${PMM_URL}/graph/api/serviceaccounts/${SA_UID}/tokens/${TOKEN_ID}" 2>/dev/null) || true
if [ "$DEL_CODE" != "200" ]; then
  echo "WARNING: could not delete token ${TOKEN_ID} (HTTP $DEL_CODE); remove it from PMM by hand"
fi
```

4nte (approved):

PR #965, opened 2026-09-12
Stop restarting PMM on every chart version bump

DrFaust92:
The PMM pod is recreated on every `helm upgrade` that carries a new chart version, even when nothing it mounts has changed. Two things in `pmm.podAnnotations` cause it:

- `checksum/config` hashes the entire rendered ConfigMap, and the manifest's `metadata.labels` include `helm.sh/chart`
- the pod annotations set `helm.sh/chart` directly

Both change on every chart version bump. This PR hashes only the ConfigMap `data` through a new `pmm.configMapOrSecretContentHash` helper, and drops the `helm.sh/chart` pod annotation — the StatefulSet itself already carries the full `pmm.labels` set, so the chart version is still visible on the object.

Rendering 1.9.2 against a bumped version now produces an identical pod template; the only differences are the `helm.sh/chart` labels on the objects themselves. A real config change still moves `checksum/config` and restarts the pod as before.

Chart version bumped to 1.9.2 per CONTRIBUTING.md (patch only). `helm lint` passes.

Note on `pmm.includeNamespace`: it detects the previous chart version by reading `helm.sh/chart` from the existing StatefulSet, trying object labels first and falling back to the pod template labels and annotations. Removing the pod annotation is safe because the object labels still carry it, which is the first entry in that chain.

The same checksum fix was recently made in the argo-cd chart (argoproj/argo-helm#4044).


DrFaust92:
Rebased onto current main — the branch was based on a stale fork and had drifted 1203 commits behind, so it was showing conflicts. Redone against pmm 1.9.1 and bumped to 1.9.2.

While rebasing I noticed `pmm.includeNamespace` reads `helm.sh/chart` off the existing StatefulSet to detect the previous chart version, including from the pod template annotations. That still works: it checks the object labels first, and those keep the label.

PR #968, opened 2026-09-14
PMM-15462 Upgrade PMM HA operators and pin the data-plane versions

theTibi:
https://perconadev.atlassian.net/browse/PMM-15462

## What

Bumps the operators pinned by `pmm-ha-dependencies` and the `pg-db` dependency of `pmm-ha`:

| | From | To |
|---|---|---|
| victoria-metrics-operator | 0.56.4 (`v0.65.0`) | 0.67.3 (`v0.74.1`) |
| altinity-clickhouse-operator | 0.25.4 | 0.27.3 |
| pg-operator | 2.8.0 | 3.1.0 |
| pg-db | 2.8.0 | 3.1.0 |

ClickHouse server and Keeper move `25.3.6` -> `25.8.28`. They are built from different
repositories (Altinity Stable rebuilds vs upstream Keeper), so they have to be moved as a pair.

## Why the data-plane pins

The chart pinned ClickHouse but inherited VictoriaMetrics from the operator and PostgreSQL from
the `pg-db` subchart. Both therefore moved as a side effect of a dependency bump, with no diff in
this chart:

- **VictoriaMetrics** was running `v1.128.0` only because that is what operator `v0.65.0` was
  built with. Operator `v0.74.1` defaults to `v1.148.0`, so this bump alone would have moved the
  metrics data plane 20 releases. Now pinned via `victoriaMetrics.version` (`v1.151.0`).
- **PostgreSQL** goes 17 -> 18, because that is the `pg-db` 3.1.0 default. That is the store
  holding `pmm-managed`, Grafana state and alert rules, so it is now `pg-db.postgresVersion: 18`
  plus an explicit image and `crVersion`, rather than inherited.

## Notes for review

- `VMCluster` spells the field `clusterVersion` and needs the `-cluster` image-tag suffix
  (`vmstorage:v1.151.0-cluster`); `VMAgent`/`VMAuth` use `componentVersion` with the bare tag.
  The value holds the bare version and each template applies what its own components need.
  A bare tag on `VMCluster` gives `ImagePullBackOff`, and neither schema validation nor
  `helm template` catches it.
- Drops `clickhouse_operator/allow_experimental_analyzer`. On ClickHouse 25.8 it is an obsolete
  alias for `enable_analyzer`, it only ever scoped the `clickhouse_operator` profile, and PMM
  Server does not set it.
- Fixes the CRD cleanup docs: operator 3.0.0 moved every Crunchy CRD to
  `upstream.pgv2.percona.com`, so the documented pattern has to match both API groups or the
  renamed CRDs survive a "full" cleanup.
- `helm upgrade` is not covered here. The PG CRD rename means Helm's install-only `crds/`
  handling leaves an existing cluster without CRDs for the new API group.

## How to test

Fresh install on a cluster with no VM/ClickHouse/PG CRDs present:

```sh
kubectl create ns pmm
# create pmm-secret - see charts/pmm-ha/README.md "Creating PMM Secret Manually"
helm install pmm-ha-dependencies charts/pmm-ha-dependencies -n pmm --wait
helm install pmm-ha charts/pmm-ha -n pmm
```

Confirm the pins took effect on the live objects:

```sh
kubectl get vmcluster -n pmm -o jsonpath='{.items[0].spec.clusterVersion}'   # v1.151.0-cluster
kubectl get vmagent,vmauth -n pmm -o jsonpath='{.items[*].spec.componentVersion}'
kubectl get perconapgcluster -n pmm -o jsonpath='{.items[0].spec.postgresVersion}'  # 18
kubectl get crds | grep -E "pgv2|crunchydata"   # only upstream.pgv2.percona.com
```

## Verified

Fresh install on a cleaned EKS cluster (50 pods): PostgreSQL `18.6.1`, ClickHouse
`25.8.28.10001.altinitystable`, VictoriaMetrics `v1.151.0`. QAN schema migrated with 3/3
ClickHouse replicas active and 0 readonly; Grafana 12.4.5 healthy against PostgreSQL 18;
VictoriaMetrics ingesting 120 targets; QAN correctly disabled for PMM's own PostgreSQL
(PMM-15301). Failover exercised for the PG primary, a ClickHouse replica, Keeper, vmstorage and
PMM Server - all recovered with PMM staying reachable.


charts/pmm-ha/values.yaml
```diff
@@ -736,6 +742,18 @@ tls:
 ## @param pg-db Configuration values passed to the pg-db dependency chart
 ##
 pg-db:
+  ## PostgreSQL major version and image, pinned deliberately.
```
JiriCtvrtka, line 745: Blocker: postgresVersion: 18 plus the PG18 image turns helm upgrade into an in-place major upgrade the operator doesn't perform. Existing instance pods come up against a PG16 data directory and die with database files are incompatible with server. Nothing guards it — the only mitigation is prose, and [README.md:651](vscode-webview://14rjts7rv4e79l3md7urt3e1ar07aa4niiht372c55p8r9n2d724/charts/pmm-ha/README.md#L651) still calls this upgrade zero-downtime. Fresh install is fine.

charts/pmm-ha-dependencies/Chart.yaml
```diff
@@ -17,11 +17,11 @@ keywords:
...
-    version: 0.56.4
+    version: 0.67.3
     repository: https://victoriametrics.github.io/helm-charts/
   - name: altinity-clickhouse-operator
-    version: 0.25.4
+    version: 0.27.3
     repository: https://helm.altinity.com
   - name: pg-operator
-    version: 2.8.0
+    version: 3.1.0
```
JiriCtvrtka, line 26: Blocker: pg-operator 3.1.0 ships its CRDs only in crds/ — I checked: no CRD template, no install hook, unlike the VM and Altinity subcharts, which do templatize theirs. Helm ignores crds/ on upgrade, so the group rename introduced in 3.0.0 never applies. The README:177 path rolls the operator binary forward while 2.8.0 CRDs are still installed, and reconciliation stops.

These compound: #2 leaves the operator unable to reconcile at exactly the moment #1 needs it, so the cluster is wedged with a one-way storage upgrade already applied and no clean rollback.

charts/pmm-ha/values.yaml
theTibi: @JiriCtvrtka 
I don't think that matters in this case , PMM HA not GA yet we can make any changes as we want even if it is not backward compatible. After we release we have to care about things like this.

JiriCtvrtka: Ok makes sense.

charts/pmm-ha-dependencies/Chart.yaml
theTibi: @JiriCtvrtka 
I have the same comment here as well, I think we can make this change now, but after GA we can't. What do you think?

JiriCtvrtka: Ok lets do it in this way.

JiriCtvrtka (approved):

